PR #134, opened 2014-07-10
Added layers to nanovg

LogicalError:
I added functionality to specify a layer when drawing. This allows the user to have more control over what is drawn on top of what. Higher layer numbers are drawn on top of lower layer numbers,


0ax1:
Are there currently any plans to realise this? I think layers would be super useful! It's actually the only thing I'm missing from nanovg :).

PR #189, opened 2014-11-21
Added nvgRoundedRectEllipse and nvgRoundedRectComplex

SmilyOrg:
We needed these to implement Flash API style logic of drawRoundRect (separate width and height for corners) and drawRoundRectComplex (separate radii for each corner), so I thought I'd PR them if you have any use for them too.


reFX-Mike:
+1

PR #210, opened 2015-03-13
Biicode support

lasote:
Hi!

My name is Luis Martinez and I write you from biicode a small start-up that’s developing a C and C++ dependency manager; just like Maven and Maven Central for Java. Our tool aims to allow our users to set-up a project with just #includes.

Biicode is based on CMake to build the projects, so its very flexible and "no-invasive" solution.

A biicode user is developing a [videogame using biicode](http://brainlesslabs.com/making-a-simple-game-nanowars-part1/) and requested us to have nanovg (and nanosvg, I will send you a PR in a minute) libraries on biicode. We think its really cool library and we decided to publish them in biicode.

https://www.biicode.com/lasote/nanovg
https://www.biicode.com/lasote/nanosvg

I think people that like your library can easy include it, GLFW and GLEW are already on biicode and included as a dependency in your project. Its tested on Windows with MinGW, Linux with GCC and OSx with Clang. Visual Studio is failing because of NVG_NOTUSED macro, I don't know if its a bug or I'm missing something.

I have an example of re-use nanovg [here](http://www.biicode.com/examples/nanovg)

In the Pull Request (and the example of use) you can see that biicode (in most cases) can work just including a CMakeLists.txt and a biicode.conf file. 
Biicode.conf declares the configuration biicode needs to build and make a library reusable:

- [requirements] section for declare dependencies to other blocks.
- [includes] Maps includes to a remote blocks, so you don't need to change your includes to depend to a remote block.

The uploaded block (we call them block, but you can think about blocks as libraries or repositories) is on lasote/nanovg, but we will glad if you want to try biicode and/or upload the library to your biicode's user. We will help you (or do it) if you need it.

If you want to know more about biicode or need help just write me!
For know a little more about biicode I recommend you to read the getting started guide [here](http://docs.biicode.com/c++/gettingstarted.html).

luis.martinez@biicode.com

Thanks!


memononen:
Hi,

Maintaining multiple build systems has been a real pain for me in the past. One of the build systems _always_ break on the next version and it's me here answering the questions I don't have answers or :)

If you wish to add and maintain biicode support, I recommend to fork.

floooh:
I'm not sure whether it is possible with biicode, but a 'non-intrusive' option might be to keep the biicode-files in a separate project, and reference nanovg from there, for instance via a git submodule. That way it is obvious for biicode users that the biicode wrapper isn't part of the original project. I've done this many times when 'fipsifying' 3rd party libraries (fips is my attempt of a cmake wrapper which seems to be partly similar to biicode in that it can import external dependecies).

For instance, here are a few examples for fipsified bfgx, glfw, glm, ...:

https://github.com/floooh/fips-bgfx/
https://github.com/floooh/fips-glfw
https://github.com/floooh/fips-glm

lasote:
@memononen I can append to the PR a .travisci and/or appveyor that checks biicode build. What do you say?
@floooh Thanks for the tip, I will take a look!

PR #221, opened 2015-05-23
Add nvgSetScissor and nvgCurrentScissor

P-i-N:
Directly getting/setting scissor rectangle, ignoring current transform
and intersections.


memononen:
I dont think this is good api design. What is the use case?

--mikko

P-i-N:
You are right, similar thing can be probably achieved with a few extra nvgSave and nvgRestore calls, so this might not be necessary after all. Will close this.

PR #226, opened 2015-06-11
Improve the use of NVGLUframebuffer.

olliwang:
The `NVGLUframebuffer` structure contains a `ctx` member but was actually never used. This commit takes advantage of the `ctx` property so there is no need to keep the reference of the context elsewhere after calling `nvgluCreateFramebuffer()`, or it would add difficulty for maintenance. As a result, the `nvgluDeleteFramebuffer()` function only needs to take one `NVGLUframebuffer` instance as the parameter.

### DIFF
--- a/example/example_fbo.c
+++ b/example/example_fbo.c
@@ -259,7 +259,7 @@ int main()
 		glfwPollEvents();
 	}
 
-	nvgluDeleteFramebuffer(vg, fb);
+	nvgluDeleteFramebuffer(fb);
 
 	nvgDeleteGL3(vg);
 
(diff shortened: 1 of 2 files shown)


olliwang:
@memononen Can you help review this pull request? Thanks. :)

memononen:
It's on my todo list :)

olliwang:
Great. Will you accept this PR? Or should I make some modifications?

memononen:
Merged, looks good. I wonder why I did not do that in the first place :) Thanks!

olliwang:
Cheers! ;)

PR #229, opened 2015-06-25
Adding nvgColorHex function

bobkocisko:
In my experience the colors that I am given from the design team are in hex format.  This patch allows those to be directly added to NanoVG

### DIFF
--- a/src/nanovg.c
+++ b/src/nanovg.c
@@ -430,6 +430,24 @@ NVGcolor nvgHSLA(float h, float s, float l, unsigned char a)
 	return col;
 }
 
+NVGcolor nvgColorHex(unsigned int color)
+{
+	// input format 0xAARRGGBB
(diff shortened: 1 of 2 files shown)


reFX-Mike:
+1

xeekworx:
This would be nice, but ...

- Isn't RGBA more common? 0xRRGGBBAA. All of the apps I use seem to follow that order; though I do realize that nanovg stores it as 0xAABBGGRR.

- The function's name should probably follow the existing color functions, eg. nvgRGB. So then maybe, nvgHexRGBA and nvgHexARGB might be appropriate.

xeekworx:
0xRRGGBB follows the html format, so then it also makes sense to ad the AA at the end. It would be the best choice and probably most expected. I store it differently internally simply so I don't have to do any conversion when I save it out to a file.

PR #232, opened 2015-07-13
Build instructions added to README.md

epitron:
I'd never run across the `premake` tool before, and there were no instructions on building the nanovg examples, so I added them.

I felt like the top of the file was a good place to put the information, since the demo is really slick -- it's good motivation to learn the API below. :)


memononen:
Excelent idea, would you mind updating the PR to give a bit more hints per platform? Here's what I did on another project: https://github.com/memononen/recastnavigation

PR #238, opened 2015-08-07
fixed type usage

plasmacel:
in some situations int and GLsizei are not the same type which leads to conversion warnings


memononen:
Thanks for the fix!

PR #245, opened 2015-09-23
Restore shader after rendering

tioover:
before rendering, save origin program id,  restore it when rendering done.
I think this fix can resolve some compatibility problem.

### DIFF
--- a/src/nanovg_gl.h
+++ b/src/nanovg_gl.h
@@ -1072,8 +1072,8 @@ static void glnvg__renderFlush(void* uptr)
 	if (gl->ncalls > 0) {
 
 		// Save origin GL Shader
-		GLint old_program;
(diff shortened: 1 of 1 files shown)


src/nanovg_gl.h
```diff
@@ -1071,9 +1071,11 @@ static void glnvg__renderFlush(void* uptr)
 
 	if (gl->ncalls > 0) {
 
-		// Setup require GL state.
+		// Save origin GL Shader
+		GLint old_program;
```
memononen: This is good change (even if the getters are often slow). Could you change `old_program` to `oldProgram` to follow the general style (I might have alternatively written `oldprog` too).

tioover: @memononen ok

PR #258, opened 2016-01-02
Switch build system generation to CMake

asarium:
I had several issues getting the examples to compile with the current premake based build system so I added a CMake based system.
The only major change was that the examples now use glad to load the OpenGL function pointers which works much better than depending on the platform to supply the right headers.

I have tested the examples on Windows and Linux. On Windows the OpenGL drivers don't support OpenGL ES so I couldn't test that on Windows but everything else works great.
On Linux OpenGL 2, 3 and OpenGL ES 2 were tested. OpenGL ES 3 failed to start but that's probably caused by missing features in the OpenGL driver.


cppfever:
Maybe who know what is problem with premake4.lua? I build nanovg, but can't examples. It's no problem for me and premake looks good. In other my projects my mingw distribution can see <GL/glew.h> and <GLFW/...> but not in examples of nanovg. What is problem?!

lilianmoraru:
@memononen Are you interested in having CMake support?
It is more portable, better integrated with tools(IDEs for example), etc... and in the C/C++ community is the most used build-system(generator), lack of support for which is mostly considered worthy of a bug report.

I see this PR removes `premake4.lua`, which definitely would be 1 reason to not be accept it(besides the other unrelated modifications) - that's why I ask if you are interested in supporting CMake?

memononen:
NanoVG is meant to be dropped into your project and the premake serves as a way to get the examples built. I've got bad experiences in the past with cmake, and at this time not willing to take the plunge to switch.

gavinhoward:
@lilianmoraru I use NanoVG as a subproject on another project that uses CMake, and it works great. Personally, even though I use CMake extensively, I don't think NanoVG should use it.

lieff:
There also new meson build system, I prefer it over cmake (better syntax, faster build). But for small project there no big difference over premake.

lilianmoraru:
@memononen @gavinhoward The idea is not to switch to it.
The idea is to have a parallel version or generate(as a few non-CMake projects choose to do) a "FindNanoVG.cmake" from the current build-system(or a CMakeLists.txt that can be added as a subdirectory).
Considering that CMake is the most used build-system in the C/C++ community, this should be considered a bug.
I personally prefer `Bazel`(sandboxed inputs, etc...) but I think it's irrelevant if this renders the project mostly unusable by other users(the majority of the C/C++ community in this case).

Again, the idea is not to replace the current build-system - of course this might imply that both of them need to be maintained when a new file is added for example(which I personally think is not a big burden).

memononen:
@lilianmoraru As I stated earlier NanoVG as a library does not need build system. You just add the relevant source files in your own project and that's it. No need for building a shared libraries or stuff like that.

For the demos, I prefer premake since it is something that I have good experience maintaining and tolerate using. I'm not going to maintain a cmake build system, and it is not needed for this project.

olliwang:
@lilianmoraru You may try building a hunter package for nanovg. That way there is no need to modify code in the nanovg repository, and it's friendly for programer who want to integrate nanovg into an existing cmake project. https://github.com/ruslo/hunter

PR #261, opened 2016-01-27
Fixed fontstash.h crashing

reFX-Mike:
Unfortunately my editor (sublime3) was configured to trim line-endings, so it looks like a lot more changes than I actually did.

Well, at least it's cleaned up now ¯\_(ツ)_/¯

### DIFF
--- a/src/nanovg.c
+++ b/src/nanovg.c
@@ -18,6 +18,9 @@
 
 #include <stdio.h>
 #include <math.h>
+#include <malloc.h>
+#include <memory.h>
+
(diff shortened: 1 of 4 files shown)


src/nanovg.c
```diff
@@ -18,6 +18,9 @@
 
 #include <stdio.h>
 #include <math.h>
+#include <malloc.h>
```
memononen: malloc.h is not a standard, should be stdlib.h instead.

memononen:
Thanks for the fix!

Issue #266, opened 2016-02-08
turning off text pixel-snapping


wtholliday:
As I understand it, nanovg's font rendering snaps in order to improve anti-aliasing.

In my app, the user can zoom in and out, so the text jumps around as it snaps. So I get fairly frequent feedback from users, who think this is a bug.

Since retina screens have such good resolution, I'd rather just turn off the snapping, but there doesn't seem to be API to do so.

How can I turn off the snapping in the code?

thanks!

MacroMachines:
This would be a useful feature, I use the app mentioned above and the text snapping is very notable and causes all sorts of little artifacts, even sometimes causing the lines to wrap at a different word at odd/even intervals so you wind up with the text jiggling all over the place when you do otherwise very smooth two finger pinch zooms

memononen:
Yes, this would be a good feature to have. I don't currently have time to implement it, but I can walk you help you out to make a PR for it.

The first thing to to would be to just remove the rounding. It is done here:
https://github.com/memononen/nanovg/blob/master/src/fontstash.h#L1134
https://github.com/memononen/nanovg/blob/master/src/fontstash.h#L1147

It could be controlled by something like this:
```C
void fonsSetPixelSnap(FONScontext* s, int state);
```
And in nanovg:
```C
void nvgTextPixelSnap(NVGcontext* ctx, int state);
```

This will lead to blurry text most of the time. In oder to solve that we could improve the font stash to include 2x2 oversampling as per: https://github.com/nothings/stb/tree/master/tests/oversample
I think that would be a better default too than pixel snapping.

guillaumechereau:
It as been a few years since the bug is open and I am also interested in disabling the snapping.  It the suggestion from @memononen still the correct way to go about it?

PR #278, opened 2016-04-19
Oversampling, Windows utf16, stbtt upgrade

itscool:
nanovg.c
- Slight modification to handle non-integer snapped text

fonstash.h
- **Support oversampling**
- Utf16 paths in fonsAddFont in Windows
- Upgraded stb_truetype.h v1.09 -> v1.11
- Questions marked with @@

Good morning! 
Oversampling :) Feedback please? Especially w.r.t questions marked with @@ in the comments.

### DIFF
--- a/src/fontstash.h
+++ b/src/fontstash.h
@@ -70,6 +70,7 @@ typedef struct FONSquad FONSquad;
 
 struct FONStextIter {
 	float x, y, nextx, nexty, scale, spacing;
+	int oversampleH, oversampleV;
 	unsigned int codepoint;
 	short isize, iblur;
 	struct FONSfont* font;
@@ -189,11 +190,18 @@ int fons__tt_getGlyphIndex(FONSttFontImpl *font, int codepoint)
 }
 
 int fons__tt_buildGlyphBitmap(FONSttFontImpl *font, int glyph, float size, float scale,
-							  int *advance, int *lsb, int *x0, int *y0, int *x1, int *y1)
+							  int oversampleH, int oversampleV,
+							  int *advance, int *lsb, int *x0, int *y0, int *x1, int *y1,
+	                          float *subX, float *subY)
 {
 	FT_Error ftError;
 	FT_GlyphSlot ftGlyph;
 	FONS_NOTUSED(scale);
+
+	// @@ How should we handle FreeType?
+
+	FONS_NOTUSED(oversampleH);
+	FONS_NOTUSED(oversampleV);
 
 	ftError = FT_Set_Pixel_Sizes(font->font, 0, (FT_UInt)(size * (float)font->font->units_per_EM / (float)(font->font->ascender - font->font->descender)));
 	if (ftError) return 0;
@@ -207,11 +215,13 @@ int fons__tt_buildGlyphBitmap(FONSttFontImpl *font, int glyph, float size, float
 	*x1 = *x0 + ftGlyph->bitmap.width;
 	*y0 = -ftGlyph->bitmap_top;
 	*y1 = *y0 + ftGlyph->bitmap.rows;
+	*subX = 0;
+	*subY = 0;
 	return 1;
 }
 
 void fons__tt_renderGlyphBitmap(FONSttFontImpl *font, unsigned char *output, int outWidth, int outHeight, int outStride,
-								float scaleX, float scaleY, int glyph)
+								float scaleX, float scaleY, int oversampleH, int oversampleV, int glyph)
 {
 	FT_GlyphSlot ftGlyph = font->font->glyph;
 	int ftGlyphOffset = 0;
@@ -220,6 +230,8 @@ void fons__tt_renderGlyphBitmap(FONSttFontImpl *font, unsigned char *output, int
 	FONS_NOTUSED(outHeight);
 	FONS_NOTUSED(scaleX);
 	FONS_NOTUSED(scaleY);
+	FONS_NOTUSED(oversampleH);
+	FONS_NOTUSED(oversampleV);
 	FONS_NOTUSED(glyph);	// glyph has already been loaded by fons__tt_buildGlyphBitmap
 
 	for ( y = 0; y < ftGlyph->bitmap.rows; y++ ) {
@@ -282,18 +294,40 @@ int fons__tt_getGlyphIndex(FONSttFontImpl *font, int codepoint)
 }
 
 int fons__tt_buildGlyphBitmap(FONSttFontImpl *font, int glyph, float size, float scale,
-							  int *advance, int *lsb, int *x0, int *y0, int *x1, int *y1)
+							  int oversampleH, int oversampleV,
+							  int *advance, int *lsb, int *x0, int *y0, int *x1, int *y1,
+							  float *subX, float *subY)
 {
 	FONS_NOTUSED(size);
 	stbtt_GetGlyphHMetrics(&font->font, glyph, advance, lsb);
-	stbtt_GetGlyphBitmapBox(&font->font, glyph, scale, scale, x0, y0, x1, y1);
+	stbtt_GetGlyphBitmapBox(&font->font, glyph, scale * oversampleH, scale * oversampleV, x0, y0, x1, y1);
+	*subX = stbtt__oversample_shift(oversampleH);
+	*subY = stbtt__oversample_shift(oversampleV);
 	return 1;
 }
 
 void fons__tt_renderGlyphBitmap(FONSttFontImpl *font, unsigned char *output, int outWidth, int outHeight, int outStride,
-								float scaleX, float scaleY, int glyph)
+								float scaleX, float scaleY, int oversampleH, int oversampleV, int glyph)
 {
-	stbtt_MakeGlyphBitmap(&font->font, output, outWidth, outHeight, outStride, scaleX, scaleY, glyph);
+	stbtt_MakeGlyphBitmapSubpixel(&font->font,
+		output, 
+		outWidth - oversampleH + 1,
+		outHeight - oversampleV + 1,
+		outStride, 
+		scaleX * oversampleH, 
+		scaleY * oversampleV,
+		0, 0, 
+		glyph);
+
+	if (oversampleH > 1)
+		stbtt__h_prefilter(output, 
+			outWidth, outHeight, outStride, 
+			oversampleH);
+
+	if (oversampleV > 1)
+		stbtt__v_prefilter(output, 
+			outWidth, outHeight, outStride, 
+			oversampleV);
 }
 
 int fons__tt_getGlyphKernAdvance(FONSttFontImpl *font, int glyph1, int glyph2)
@@ -353,7 +387,12 @@ struct FONSglyph
 	int next;
 	short size, blur;
 	short x0,y0,x1,y1;
-	short xadv,xoff,yoff;
+	int oversampleV, oversampleH;
+
+	// @@ Using shorts and mul/div by 10 is probably overkill at 
+	//    this point, unless we REALLY need to save that memory?
+
+	float xadv,xoff,yoff,xoff2,yoff2;
 };
 typedef struct FONSglyph FONSglyph;
 
@@ -382,6 +421,8 @@ struct FONSstate
 	unsigned int color;
 	float blur;
 	float spacing;
+	int oversampleH;
+	int oversampleV;
 };
 typedef struct FONSstate FONSstate;
 
@@ -781,6 +822,12 @@ void fonsSetFont(FONScontext* stash, int font)
 	fons__getState(stash)->font = font;
 }
 
+void fonsSetOversample(FONScontext* stash, int oversampleH, int oversampleV)
+{
+	fons__getState(stash)->oversampleH = oversampleH;
+	fons__getState(stash)->oversampleV = oversampleV;
+}
+
 void fonsPushState(FONScontext* stash)
 {
 	if (stash->nstates >= FONS_MAX_STATES) {
@@ -811,6 +858,8 @@ void fonsClearState(FONScontext* stash)
 	state->font = 0;
 	state->blur = 0;
 	state->spacing = 0;
+	state->oversampleH = 2;
+	state->oversampleV = 2;
 	state->align = FONS_ALIGN_LEFT | FONS_ALIGN_BASELINE;
 }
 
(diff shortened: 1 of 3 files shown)


src/fontstash.h
```diff
@@ -189,12 +190,19 @@ int fons__tt_getGlyphIndex(FONSttFontImpl *font, int codepoint)
...
-							  int *advance, int *lsb, int *x0, int *y0, int *x1, int *y1)
+							  int oversampleH, int oversampleV,
+							  int *advance, int *lsb, int *x0, int *y0, int *x1, int *y1,
+	                          float *subX, float *subY)
 {
 	FT_Error ftError;
 	FT_GlyphSlot ftGlyph;
 	FONS_NOTUSED(scale);
 
+	// @@ How should we handle FreeType?
```
memononen: We should render at higher resolution and box blur the resulting glyph. The larger resolution should come from oversample, and it's also the box blur size.

src/fontstash.h
```diff
@@ -282,18 +294,40 @@ int fons__tt_getGlyphIndex(FONSttFontImpl *font, int codepoint)
...
+	*subY = stbtt__oversample_shift(oversampleV);
 	return 1;
 }
 
 void fons__tt_renderGlyphBitmap(FONSttFontImpl *font, unsigned char *output, int outWidth, int outHeight, int outStride,
-								float scaleX, float scaleY, int glyph)
+								float scaleX, float scaleY, int oversampleH, int oversampleV, int glyph)
 {
-	stbtt_MakeGlyphBitmap(&font->font, output, outWidth, outHeight, outStride, scaleX, scaleY, glyph);
+	stbtt_MakeGlyphBitmapSubpixel(&font->font,
```
memononen: Put more params on one line to follow the coding style, i.e. break after outStride.

src/fontstash.h
```diff
@@ -282,18 +294,40 @@ int fons__tt_getGlyphIndex(FONSttFontImpl *font, int codepoint)
...
+		outWidth - oversampleH + 1,
+		outHeight - oversampleV + 1,
+		outStride, 
+		scaleX * oversampleH, 
+		scaleY * oversampleV,
+		0, 0, 
+		glyph);
+
+	if (oversampleH > 1)
+		stbtt__h_prefilter(output,
```
memononen: all params on same line

src/fontstash.h
```diff
@@ -282,18 +294,40 @@ int fons__tt_getGlyphIndex(FONSttFontImpl *font, int codepoint)
...
+		0, 0, 
+		glyph);
+
+	if (oversampleH > 1)
+		stbtt__h_prefilter(output, 
+			outWidth, outHeight, outStride, 
+			oversampleH);
+
+	if (oversampleV > 1)
+		stbtt__v_prefilter(output,
```
memononen: ditto

src/fontstash.h
```diff
@@ -1007,20 +1079,34 @@ static void fons__blur(FONScontext* stash, unsigned char* dst, int w, int h, int
...
 	unsigned char* bdst;
 	unsigned char* dst;
 
 	if (isize < 2) return NULL;
 	if (iblur > 20) iblur = 20;
-	pad = iblur+2;
+	if (oversampleH < 1) oversampleH = 1;
+	if (oversampleV < 1) oversampleV = 1;
+
+	// Only use extra 1px "leak" padding if blurring (otherwise it breaks oversampling)
```
memononen: We always need 2px around the actual pixel data to render the glyphs correctly. One pixel inside the rendered quad and one pixel outside. The reason is that otherwise rotated glyphs will pick up neighbour data.

src/fontstash.h
```diff
@@ -1007,20 +1079,34 @@ static void fons__blur(FONScontext* stash, unsigned char* dst, int w, int h, int
...
+	// To compensate, we disable oversampling if blurred (doesn't make sense to combine anyway)
```
memononen: We should use oversampling for blur radius < 2, otherwise the transition may not be smooth.

src/fontstash.h
```diff
@@ -353,7 +387,12 @@ struct FONSglyph
 	int next;
 	short size, blur;
 	short x0,y0,x1,y1;
-	short xadv,xoff,yoff;
+	int oversampleV, oversampleH;
```
memononen: use `char` or `short` here.

src/fontstash.h
```diff
@@ -353,7 +387,12 @@ struct FONSglyph
...
+
+	// @@ Using shorts and mul/div by 10 is probably overkill at 
+	//    this point, unless we REALLY need to save that memory?
+
+	float xadv,xoff,yoff,xoff2,yoff2;
```
memononen: floats ok, put these floats right after `next`

src/fontstash.h
```diff
@@ -781,6 +822,12 @@ void fonsSetFont(FONScontext* stash, int font)
 	fons__getState(stash)->font = font;
 }
 
+void fonsSetOversample(FONScontext* stash, int oversampleH, int oversampleV)
+{
+	fons__getState(stash)->oversampleH = oversampleH;
```
memononen: We probably should validate oversampling here. I.e. make sure it is between 1 and 16 or something like that.

src/fontstash.h
```diff
@@ -811,6 +858,8 @@ void fonsClearState(FONScontext* stash)
 	state->font = 0;
 	state->blur = 0;
 	state->spacing = 0;
+	state->oversampleH = 2;
```
memononen: default should match old behavior which is 1

src/fontstash.h
```diff
@@ -1113,52 +1207,42 @@ static void fons__getQuad(FONScontext* stash, FONSfont* font,
...
+	float xoff, yoff, xoff2, yoff2;
+	short x0, y0, x1, y1;
 	if (prevGlyphIndex != -1) {
 		float adv = fons__tt_getGlyphKernAdvance(&font->font, prevGlyphIndex, glyph->index) * scale;
-		*x += (int)(adv + spacing + 0.5f);
+		*x += adv + spacing;
 	}
 
-	// Each glyph has 2px border to allow good interpolation,
+	// Each blurred glyph has 2px border to allow good interpolation,
```
memononen: As per comment elsewhere, always use 2px padding, change comment too.

src/fontstash.h
```diff
@@ -1113,52 +1207,42 @@ static void fons__getQuad(FONScontext* stash, FONSfont* font,
...
-		q->y0 = ry;
-		q->x1 = rx + x1 - x0;
-		q->y1 = ry - y1 + y0;
-
-		q->s0 = x0 * stash->itw;
-		q->t0 = y0 * stash->ith;
-		q->s1 = x1 * stash->itw;
-		q->t1 = y1 * stash->ith;
-	}
+	// @@ Do we want option to calculate pixel-snapped coords?
```
memononen: Yes. if oversampling == 1 then we should snap at that dimension.

src/fontstash.h
```diff
@@ -1113,52 +1207,42 @@ static void fons__getQuad(FONScontext* stash, FONSfont* font,
...
+		q->y1 = *y + yoff2;
+	else
+		q->y1 = *y - yoff2;
+
+	q->s0 = x0 * stash->itw;
+	q->t0 = y0 * stash->ith;
+	q->s1 = x1 * stash->itw;
+	q->t1 = y1 * stash->ith;
 
-	*x += (int)(glyph->xadv / 10.0f + 0.5f);
```
memononen: These 0.5s sprinkled all over are important for good pixel snapping.

src/fontstash.h
```diff
@@ -1113,52 +1207,42 @@ static void fons__getQuad(FONScontext* stash, FONSfont* font,
 						   int prevGlyphIndex, FONSglyph* glyph,
 						   float scale, float spacing, float* x, float* y, FONSquad* q)
 {
-	float rx,ry,xoff,yoff,x0,y0,x1,y1;
-
+	float xoff, yoff, xoff2, yoff2;
+	short x0, y0, x1, y1;
 	if (prevGlyphIndex != -1) {
 		float adv = fons__tt_getGlyphKernAdvance(&font->font, prevGlyphIndex, glyph->index) * scale;
-		*x += (int)(adv + spacing + 0.5f);
```
memononen: Should use the old code if horiz oversample = 1

src/nanovg.c
```diff
@@ -2545,11 +2545,13 @@ int nvgTextBreakLines(NVGcontext* ctx, const char* string, const char* end, floa
 					breakMaxX = 0.0;
 				}
 			} else {
-				float nextWidth = iter.nextx - rowStartX;
```
memononen: There are two bounds on each glyph, the "logical" i.e. what the advance is based on and the the actual geometric bounds. You have to be careful to make sure you always use the same ones, else you get hard to find bugs. This code was particularly nasty to get working.

This piece of code is dealing with the "logical" units of the glyphs. Unfortunately I cannot recall all the details why I chose to do it that way, but I think it had to do with some obvious text wrapping without a good reason.

Did you change this because of some bug?

memononen:
Added some comments. I think you should add API to the NanoVG too to set the text pixel snapping. See https://github.com/memononen/nanovg/issues/266

src/fontstash.h
itscool: I just want to be sure FreeType doesn't already do something equivalent in its default sub-pixel renderer or as an option.

src/fontstash.h
itscool: I'm sorry, I do not understand. If there is a 1 pixel padding **on all sides per glyph**, effectively giving 2 empty pixels pad between cached glyphs. As it currently works, with 1 pixel outside, and 1 pixel inside, on all sides, per glyph, there is effectively 4 pixels between glyphs. No?

itscool: See, with oversampling, having an extra pixel padding inside doesn't work, because we would also need to multiply that 1 pixel padding by the oversampling scale. We end up with a whole lot of padding and unused texture space to enable oversampling...

src/fontstash.h
itscool: If oversampleV/H are `char` alignment should be fine, but yeah better to move them.

src/fontstash.h
itscool: Default on my displays was pretty blurry. You sure?

src/fontstash.h
itscool: Of course

src/nanovg.c
itscool: Yes. Per the comment, sub-pixel positioning brought out a bug where the advance would be inside the last glyph's max x. 

So the code would calculate bounding box based on glyph dimensions, but calculate line break based on advance, and my last character or word would not be displayed because of line break, when rendering and clipping within bounding box.

src/fontstash.h
memononen: It's a good idea to check that. I have not used FT for a long time, though.

src/fontstash.h
memononen: It should be really crisp when pixel snapping is on. Can you post a screenshot?

src/fontstash.h
memononen: I remember adding that second padding pixel because rotated text would not work otherwise. I recommend to change the padding to 1 and try rotated text. Test snapped, oversampled, and blurred. If it works, then the problem was somewhere else and we can keep just the 1px padding. If it does not, then we'll need more padding (yeah, sucks).

src/nanovg.c
memononen: The comment does not make sense at it refers to some problem which is not described. Let's not focus on that yet.

So your use case was that you first call `nvgTextBoxBounds()` to effectively figure out how many lines there are and then draw using `nvgTextBox()` based on the returned bounds? If that is the case, it is exactly what I tried to solve with the current implementation.

That case is also really hard to get right because of the two effective bounds in fonts. For example big O or cursive small y can be optically aligned, so even if you ask to render them at certain location they actually start few pixels "before". In that case if you call text bounds and use the glyph bounds (geometry) to calculate stuff and render again (position in "logical" coordinates), you'll get offset results. It is complicated and annoying. That's the reason the rows have `width`(logical) as well as `minx`, `maxx` (geometry).

Can you post the code which you trigger the buggy case use? Let's see if there's some assumption there which fights against the two coordinate system stuff.

itscool: Hmm I don't think my communication is good here. I'll show you the issue when I get some time.

src/fontstash.h
itscool: That can work.

src/fontstash.h
itscool: Hmm. I'm thinking of possible ways I could have misused it. Will write back and/or post screenshot in the next few days (hopefully).

src/fontstash.h
itscool: I have some thoughts on this. Maybe we'll save this and tackle it last.

itscool:
How does this work? It looks like I just commit to my own branch and it updates here?

memononen:
When you push changes to your branch, it should update the PR.

src/fontstash.h
itscool: Regarding convention, is there a method to the single line `if` statements? For example, sometimes `return` comes on same line as the conditional, sometimes it is on the next line indented.

src/fontstash.h
itscool: Inconsistent... would you like other fonsSet* functions to include validation? Or is there a good reason to only use here? **or** perhaps oversampling should simply be on/off (2x2 gives good results, after that can be questionable). Perhaps it should be automatic if pixel snapping is off? What do you think?

src/fontstash.h
itscool: - [ ] Done

src/fontstash.h
itscool: - [x] Done

src/fontstash.h
itscool: - [x] Done

src/fontstash.h
itscool: - [x] Done

src/fontstash.h
itscool: - [x] Done

src/fontstash.h
itscool: Also in #266 you actually mentioned 2x2 oversample would be a better default... still waiting to get back to you on the blur issue though.

src/fontstash.h
itscool: I think this is my biggest question right now to move forward. It also can address FreeType issue. IF we decide to just have pixel-snapping on/off (off giving 2x2 oversampling), then pixel snapping off can just revert to sub-pixel rendering in FreeType (which works very well on its own). So we end up with sub-pixel, or pixel-snapping, and it looks good with both STBTT and FreeType. I really think this is good...

src/fontstash.h
memononen: Obvious ifs with early outs can be on same line, generally put stuff on a new line.

src/fontstash.h
memononen: I think a toggle would be just fine as per #266. Oversample amount could be just a define. I don't think FT subpixel rendering will work directly here.

What I'm expect the stb_tt to do is to render the glyph twice the size, and box blur it. Then when rendering we render the glyph half the size. This effectively gives us "more samples" when rendering. If I understand correctly, FT sub pixel rendering just renders the glyph, but does not snap to pixels. So we should do double size rendering and blur on FT too.

src/fontstash.h
memononen: Technically, I think it would be, but it could also be surprising. Let's be backwards compatible on this one.

src/fontstash.h
memononen: Re: validation, I tend to add some when I think that there's a good sane range and when values outside that range can have big negative impact elsewhere (i.e. oversample = 1000).

PR #284, opened 2016-05-24
Fix int-to-float conversion warning in nvgTextBreakLines

JinShil:


JinShil:
Sorry, it appears to be a problem with my compiler.

PR #295, opened 2016-08-02
Fixing clang compiler warnings.

nhojb:
Just a few small changes to silence warnings when compiling in Xcode 7.3.1 (llvm 7.3.0).

Note that I'm also using the freetype2 font support, hence the additional #ifdef to silence unused function warnings.


memononen:
Thanks for the fixes!

nhojb:
Thanks for accepting them :-)

PR #297, opened 2016-08-05
Add support for bgfx

olliwang:
This commit updates the `nvgBeginFrame()` function to forward the received `devicePixelRatio` parameter to the `renderViewport()` function so it is possible to configure the environemnt correctly when using `bgfx` backend.

The current implementation of the `renderViewport()` function in bgfx backend is implemented in this line: https://github.com/bkaradzic/bgfx/blob/7de947c7c08acdc597bc0bd3f67f43e2bafd3760/examples/common/nanovg/nanovg_bgfx.cpp#L553 

However, the `bgfx::setViewRect()` function won't work properly if the passed width and height are not multiplied to the `devicePixelRatio`, and because `devicePixelRatio` is kept in `NVGcontext` and the `NVGcontext` object is only visible in the `nanovg.c` file. It seems that passing the `devicePixelRatio` parameter to the `renderViewport()` function is the only way to make `bgfx` backend works properly.

### DIFF
--- a/src/nanovg.c
+++ b/src/nanovg.c
@@ -295,7 +295,7 @@ void nvgBeginFrame(NVGcontext* ctx, int windowWidth, int windowHeight, float dev
 
 	nvg__setDevicePixelRatio(ctx, devicePixelRatio);
 
-	ctx->params.renderViewport(ctx->params.userPtr, windowWidth, windowHeight);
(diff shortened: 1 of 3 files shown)


memononen:
Thanks for the improvement, and good luck with the metal port :)

olliwang:
Thanks. ;)

PR #299, opened 2016-08-16
Support fallback fonts

satoren:
I want to render an Emoji in same context.

added two functions.
```
int nvgAddFallbackFontId(NVGcontext* ctx,int baseFont,int fallbackFont);
int nvgAddFallbackFont(NVGcontext* ctx,const char* baseFont,const char* fallbackFont);
```
example
```
data->fontNormal = nvgCreateFont(vg, "sans", "../example/Roboto-Regular.ttf");
data->fontEmoji = nvgCreateFont(vg, "emoji", "../example/NotoEmoji-Regular.ttf");
//If not available codepoint in "sans" , fallback to "emoji"
nvgAddFallbackFontId(vg,data->fontNormal,data->fontEmoji);
```
Screenshot

PR #301, opened 2016-08-20
Support fallback fonts (code refactoring for #299)

olliwang:
This PR fixes the formats as suggested in https://github.com/memononen/nanovg/pull/299 and fixes the bug that the original implementation only works for `stb_truetype` but not `freetype`.


src/fontstash.h
```diff
@@ -1041,6 +1058,18 @@ static FONSglyph* fons__getGlyph(FONScontext* stash, FONSfont* font, unsigned in
 	// Could not find glyph, create it.
 	scale = fons__tt_getPixelHeightScale(&font->font, size);
 	g = fons__tt_getGlyphIndex(&font->font, codepoint);
+	// Try to find the glyph in fallback fonts.
+	if (g == 0)
+	{
+		for (i = 0; i < font->nfallbacks; ++i)
+		{
+			fallbackGlyph = fons__getGlyph(stash, stash->fonts[font->fallbacks[i]], codepoint, isize, iblur);
+			if (fallbackGlyph != NULL)
```
olliwang: The original implementation checks if `glyph->index != 0`. However, this only works for `stb_truetype`. To make it work on both `stb_truetype` and `freetype`, we can only check if the fallback glyph is NULL. Checking `glyph->index` causes issue when using `freetype` backend.

src/fontstash.h
```diff
@@ -1041,6 +1058,18 @@ static FONSglyph* fons__getGlyph(FONScontext* stash, FONSfont* font, unsigned in
 	// Could not find glyph, create it.
 	scale = fons__tt_getPixelHeightScale(&font->font, size);
 	g = fons__tt_getGlyphIndex(&font->font, codepoint);
+	// Try to find the glyph in fallback fonts.
+	if (g == 0)
+	{
+		for (i = 0; i < font->nfallbacks; ++i)
+		{
+			fallbackGlyph = fons__getGlyph(stash, stash->fonts[font->fallbacks[i]], codepoint, isize, iblur);
```
memononen: You can do:
```
FONSglyph* fallbackGlyph = ...
```
variable definitions are ok at the beginning of any scope.

src/fontstash.h
```diff
@@ -755,6 +760,17 @@ static FONSstate* fons__getState(FONScontext* stash)
 	return &stash->states[stash->nstates-1];
 }
 
+int fonsAddFallbackFont(FONScontext* stash, int base, int fallback)
+{
+	FONSfont* baseFont = stash->fonts[base];
+	if (baseFont->nfallbacks < FONS_MAX_FALLBACKS)
+	{
```
memononen: Braces on the same line with if. Please check other changes too.

src/fontstash.h
memononen: Checking for null only does not sound quite right (but surely should check null too). fons__getGlyph should return null only when the atlas is full. The code is currently made so that even if we hit 0 index (i.e. no glyph found), we still create cache entry for that to avoid potentially costly lookups later. Both freetype and stbtt should return 0 for missing glyphs.

olliwang: That's a bit weird. I tried to render only the emoji character using the same codebase with different backends. The codepoint of the emoji is `127881`. When I ran my app with `stb_truetype` backend, the glyph index returns `298`, and the emoji is rendered as expected. However, after I changed to `freetype` backend. I can still see the codepoint is `127811`, but the glyph index also returns `0` and never render the emoji. I have no idea why this happened.

olliwang:
I thought I just found why checking the glyph index doesn't work for me. I found a very weird situation on my iOS environment. The `g` variable is reset after calling `fons__tt_buildGlyphBitmap()`. 

      printf("before: %d\n", g);
      int glyph_index = g;
      fons__tt_buildGlyphBitmap(&font->font, g, size, scale, &advance, &lsb, &x0, &y0, &x1, &y1);
      printf("after: %d\n", g);
      printf("copied value: %d\n", glyph_index);

This prints

    before: 298
    after: 0
    copied value: 298

This is the reason why I always got index `0` when using `freetype` backend.

olliwang:
Don't exact know why but I make the weird situation gone (as described above) by declaring the `g` variable separately. Now the latest commit fixes the formats and works on both backends.

memononen:
@olliwang that usually means that something is overriding the stack. For example changin the order in which the variables are declared can have same effect. Could you try to add a trigger breakpoint to check when the variable changes?
http://stackoverflow.com/questions/18150221/how-do-i-have-a-breakpoint-get-triggered-if-an-instance-variable-in-the-class-ha
Now maybe the `advance` gets overridden. Also, camelCase not under_score in variable names :)

olliwang:
@memononen Thanks a lot. Putting `g` behind `advance` did solve this issue. I've updated the code. 👍 (PS. I couldn't get the breakpoint work on a header file)

memononen:
@olliwang that solves the problem, but there's still stack overflow. 

Can you try to change the following in Freetype version:
```C
int fons__tt_buildGlyphBitmap(FONSttFontImpl *font, int glyph, float size, float scale,
							  int *advance, int *lsb, int *x0, int *y0, int *x1, int *y1)
{
	FT_Error ftError;
	FT_GlyphSlot ftGlyph;
	FT_Fixed advFixed;
	FONS_NOTUSED(scale);

	ftError = FT_Set_Pixel_Sizes(font->font, 0, (FT_UInt)(size * (float)font->font->units_per_EM / (float)(font->font->ascender - font->font->descender)));
	if (ftError) return 0;
	ftError = FT_Load_Glyph(font->font, glyph, FT_LOAD_RENDER);
	if (ftError) return 0;
	ftError = FT_Get_Advance(font->font, glyph, FT_LOAD_NO_SCALE, advFixed);
	if (ftError) return 0;
	ftGlyph = font->font->glyph;
	*advance = (int)advFixed;
	*lsb = (int)ftGlyph->metrics.horiBearingX;
	*x0 = ftGlyph->bitmap_left;
	*x1 = *x0 + ftGlyph->bitmap.width;
	*y0 = -ftGlyph->bitmap_top;
	*y1 = *y0 + ftGlyph->bitmap.rows;
	return 1;
}
```
I suspect that `FT_Fixed` is 64bit value (unsigned long), and passing int pointer there stores more than it should messing up the stack.

olliwang:
@memononen  You're right. Your code fixes the stash overflow issue and I changed the order of `g` back. Thanks. 👍

memononen:
Great 👍  Merged, and thanks for picking up the fixes and PR updates!

Issue #302, opened 2016-08-22
Another stack overflow when using freetype?


olliwang:
It seems something is wrong when determining the glyph `advance` in freetype. Maybe another stack overflow? The left picture is rendered using `stb_truetype` and the right one is rendered using `freetype`, and happens after the recent change (#301) in `fons__tt_buildGlyphBitmap()`.

![screen shot 2016-08-22 at 23 22 42](https://cloud.githubusercontent.com/assets/76374/17860358/d97036a6-68bf-11e6-84b7-6404aff517b4.png)

However, though in this case the old implementation of `fons__tt_buildGlyphBitmap()` makes this situation gone, but it did happen again when trying to render a different text.

After several hours of investigating. I've found that `FT_Get_Kerning()` returns a negative number of the last character(T) while the others(NOR) return `0`. But still have no idea how to solve this issue. @memononen Can you give me some advices? Thanks.

memononen:
Maybe the Freetype versio of `fons__tt_getGlyphKernAdvance` should convert from it's 26.6 fixed point to an integer pixel value.

Can you try this:
```C
int fons__tt_getGlyphKernAdvance(FONSttFontImpl *font, int glyph1, int glyph2)
{
	FT_Vector ftKerning;
	FT_Get_Kerning(font->font, glyph1, glyph2, FT_KERNING_DEFAULT, &ftKerning);
	return (int)((ftKerning.x + 32) >> 6);  // Round up and convert to integer
}
```
In case of R and T negative kerning sounds correct.

olliwang:
@memononen Wow! That's the reason why. You just saved my day. I opened PR #303 to fix that. Thank you very much. 👍

PR #303, opened 2016-08-23
Fixes the kerning value for freetype.

olliwang:
This commit fixes issue #302 that the kerning value returned by freetype is not converted to integer properly.


memononen:
Thanks for the PR!

Issue #214, opened 2015-03-31
removing discard from the fragment shader significantly improves performance on iOS


wtholliday:
Before:

![before](https://cloud.githubusercontent.com/assets/841144/6924316/9a90b34a-d78c-11e4-8d4c-42c56cf37d87.png)

After:

![after](https://cloud.githubusercontent.com/assets/841144/6924323/a1eb90c4-d78c-11e4-8a6a-4afa667b9938.png)

I did notice some rendering artifacts for self-intersecting paths:

![img_0052](https://cloud.githubusercontent.com/assets/841144/6924440/64414fec-d78d-11e4-88a9-21786d70744b.PNG)

vs.

![img_0053](https://cloud.githubusercontent.com/assets/841144/6924446/6bf1c3a2-d78d-11e4-9baa-a89befa7e96e.PNG)

How should I fix the artifacts? Should there be an option exposed to turn off discard (it's a no-no on iOS)?

dougbinks:
It should be possible to set the alpha value to 0.0 instead of using discard to make the pixel completely transparent. Discard is indeed an issue on many mobile configurations, so this should likely be done for all OpenGL ES configurations.

wtholliday:
@dougbinks, surprisingly, setting the alpha to 0.0 doesn't fix the artifacts. In fact it creates a new problem with strokes:

![img_0054](https://cloud.githubusercontent.com/assets/841144/6925681/e66f80da-d795-11e4-84b0-d9d34a8747b8.PNG)

I did the following:

```
    "#ifdef EDGE_AA\n"
    "	if (strokeAlpha < strokeThr) { result.a = 0.0; };\n"
    "#endif\n"
```

dougbinks:
The blend function is glBlendFunc(GL_ONE, GL_ONE_MINUS_SRC_ALPHA); so this is likely due to depth or stencil testing issues and so not easily solvable though Mikko might be able to come up with a solution.

Since the discard is only on the EDGE_AA path, I would not use the NVG_ANTIALIAS if you want performance.

wtholliday:
Good thought about the stenciling. Turning off NVG_STENCIL_STROKES gets rid of the artifacts, but of course then self-intersecting paths are drawn differently (which I can live with).

memononen:
Good catch! I wonder if splitting the shaders would make it faster on mobile too? So that if one shader uses discard, all shaders don't pay the penalty?

[...]

PR #305, opened 2016-08-31
Update nanovg_gl.h

swittk:
Changing from discarding the result ("discard;") to assigning the result variable with vec4(0,0,0,0) appears to drastically improve performance on iOS.

(FPS running NanoVG Demo on iPhone 5, iOS 7.0 : From 13 FPS on average to around 40 FPS)


wtholliday:
This breaks NVG_STENCIL_STROKES. See https://github.com/memononen/nanovg/issues/214

Issue #306, opened 2016-09-01
You have a problem in "nanovg_gl.h"


ProgrammingLife:
You have a problem in /nanovg/src/nanovg_gl.h:1099:1, error: "Control may reach end of non-void function". You need to define default case if no one of those "if-sections" was reached. What the default value define there? :)

PR #307, opened 2016-09-04
Fallbacks to default composite operation when invalid one is specified.

### LINKED ISSUE #300: glnvg_convertBlendFuncFactor: control may reach end of non-void function
The recent PR #298 triggers a warning with AppleClang in the function `glnvg_convertBlendFuncFactor`:

```
In file included from third_party/nanovg/nanovg_gl2.c:4:
third_party/nanovg/nanovg_gl.h:1099:1: warning: 
      control may reach end of non-void function [-Wreturn-type]
```

There just needs to be a default return statement.

olliwang:
This commit fixes #300 and #306 that “control may reach end of non-void function” error occurs in `glnvg_convertBlendFuncFactor()`. The function now returns `GL_INVALID_ENUM` for invalid blend factor and if that happens, `glnvg__blendCompositeOperation()` uses default source-over composite operation.


memononen:
Thanks for the fix!

PR #311, opened 2016-09-11
Rename nvglImageFromHandleGL2 -> nvglImageHandleGL2

dimsuz:
There was no `nvglImageFromHandleGL2` implementation, only declaration, it must have been
an error during some refactoring.

This commit renames `nvglImageFromHandleGL2` to `nvglImageHandleGL2` so now this declaration matches definition at the bottom of the file and is named similarly to other GL* functions with same name


memononen:
Thanks for the fix!

PR #316, opened 2016-09-18
nvgRoundedRectEx for varying corner radii

xeekworx:
Now each corner of the rounded rectangle can have it's own radius.

### DIFF
--- a/src/nanovg.c
+++ b/src/nanovg.c
@@ -2119,22 +2119,30 @@ void nvgRect(NVGcontext* ctx, float x, float y, float w, float h)
 
 void nvgRoundedRect(NVGcontext* ctx, float x, float y, float w, float h, float r)
 {
-	if (r < 0.1f) {
-		nvgRect(ctx, x,y,w,h);
+	nvgRoundedRectEx(ctx, x, y, w, h, r, r, r, r);
+}
+
+void nvgRoundedRectEx(NVGcontext* ctx, float x, float y, float w, float h, float r1, float r2, float r3, float r4)
+{
+	if(r1 < 0.1f && r2 < 0.1f && r3 < 0.1f && r4 < 0.1f) {
+		nvgRect(ctx, x, y, w, h);
 		return;
 	}
 	else {
-		float rx = nvg__minf(r, nvg__absf(w)*0.5f) * nvg__signf(w), ry = nvg__minf(r, nvg__absf(h)*0.5f) * nvg__signf(h);
+		float r1x = nvg__minf(r1, nvg__absf(w)*0.5f) * nvg__signf(w), r1y = nvg__minf(r1, nvg__absf(h)*0.5f) * nvg__signf(h);
+		float r2x = nvg__minf(r2, nvg__absf(w)*0.5f) * nvg__signf(w), r2y = nvg__minf(r2, nvg__absf(h)*0.5f) * nvg__signf(h);
+		float r3x = nvg__minf(r3, nvg__absf(w)*0.5f) * nvg__signf(w), r3y = nvg__minf(r3, nvg__absf(h)*0.5f) * nvg__signf(h);
+		float r4x = nvg__minf(r4, nvg__absf(w)*0.5f) * nvg__signf(w), r4y = nvg__minf(r4, nvg__absf(h)*0.5f) * nvg__signf(h);
 		float vals[] = {
-			NVG_MOVETO, x, y+ry,
-			NVG_LINETO, x, y+h-ry,
-			NVG_BEZIERTO, x, y+h-ry*(1-NVG_KAPPA90), x+rx*(1-NVG_KAPPA90), y+h, x+rx, y+h,
-			NVG_LINETO, x+w-rx, y+h,
-			NVG_BEZIERTO, x+w-rx*(1-NVG_KAPPA90), y+h, x+w, y+h-ry*(1-NVG_KAPPA90), x+w, y+h-ry,
-			NVG_LINETO, x+w, y+ry,
-			NVG_BEZIERTO, x+w, y+ry*(1-NVG_KAPPA90), x+w-rx*(1-NVG_KAPPA90), y, x+w-rx, y,
-			NVG_LINETO, x+rx, y,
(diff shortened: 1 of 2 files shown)


src/nanovg.c
```diff
@@ -2119,22 +2119,30 @@ void nvgRect(NVGcontext* ctx, float x, float y, float w, float h)
 
 void nvgRoundedRect(NVGcontext* ctx, float x, float y, float w, float h, float r)
 {
-	if (r < 0.1f) {
-		nvgRect(ctx, x,y,w,h);
+	nvgRoundedRectEx(ctx, x, y, w, h, r, r, r, r);
+}
+
+void nvgRoundedRectEx(NVGcontext* ctx, float x, float y, float w, float h, float r1, float r2, float r3, float r4)
```
btipling: These look like vectors, why not create a vector struct type? You could reduce the argument spam.

btipling: I suppose in the case of the round corners, I wonder if you would ever want a separate radius for each of the corners.

src/nanovg.c
```diff
@@ -2119,22 +2119,30 @@ void nvgRect(NVGcontext* ctx, float x, float y, float w, float h)
...
+	if(r1 < 0.1f && r2 < 0.1f && r3 < 0.1f && r4 < 0.1f) {
+		nvgRect(ctx, x, y, w, h);
 		return;
 	}
 	else {
-		float rx = nvg__minf(r, nvg__absf(w)*0.5f) * nvg__signf(w), ry = nvg__minf(r, nvg__absf(h)*0.5f) * nvg__signf(h);
+		float r1x = nvg__minf(r1, nvg__absf(w)*0.5f) * nvg__signf(w), r1y = nvg__minf(r1, nvg__absf(h)*0.5f) * nvg__signf(h);
+		float r2x = nvg__minf(r2, nvg__absf(w)*0.5f) * nvg__signf(w), r2y = nvg__minf(r2, nvg__absf(h)*0.5f) * nvg__signf(h);
+		float r3x = nvg__minf(r3, nvg__absf(w)*0.5f) * nvg__signf(w), r3y = nvg__minf(r3, nvg__absf(h)*0.5f) * nvg__signf(h);
+		float r4x = nvg__minf(r4, nvg__absf(w)*0.5f) * nvg__signf(w), r4y = nvg__minf(r4, nvg__absf(h)*0.5f) * nvg__signf(h);
```
btipling: This code looks a bit repetitive I'd make a separate function or iterate over an array of values.

src/nanovg.c
xeekworx: btipling ... This is C not C++; otherwise I would have just made the existing function have defaults or made the other version of it with the same name. Since it's C the new function had to also be named differently. Of course that also means we don't have vectors in C.

btipling: structs work in C also

src/nanovg.c
xeekworx: Probably so. This was clearly a lazy copy and paste on my part.

xeekworx: I thought about it and well, I just wouldn't prefer having an array as a parameter because that would probably mean I can't use this function without another line of code to declare that array. That also means iterating to setup the rx and ry values isn't very doable.  I'm happy with it as is. Maybe if this were C++ there would absolutely be much more possibilities. I could replace those operations with a function, but I would still be left with the same lines of code with maybe an inch of columns saved. In the end would just be about reducing code rather than optimization of actual execution.

src/nanovg.c
memononen: `Ex` for "extra" functions is bad habit :) How about ´nvgRoundedRectVarying` instead?
The parameters should be at least a bit self documenting. They should be in CSS order: top-left, top-right, bottom-right, and bottom-left. So something like:
```C
void nvgRoundedRectVarying(NVGcontext* ctx, float x, float y, float w, float h, float radTopLeft, float radTopRight, float radBottomRight, float radBottomLeft)
```

src/nanovg.c
```diff
@@ -2119,22 +2119,30 @@ void nvgRect(NVGcontext* ctx, float x, float y, float w, float h)
...
+	nvgRoundedRectEx(ctx, x, y, w, h, r, r, r, r);
+}
+
+void nvgRoundedRectEx(NVGcontext* ctx, float x, float y, float w, float h, float r1, float r2, float r3, float r4)
+{
+	if(r1 < 0.1f && r2 < 0.1f && r3 < 0.1f && r4 < 0.1f) {
+		nvgRect(ctx, x, y, w, h);
 		return;
 	}
 	else {
```
memononen:
```C
if (...) {
} else {
}
```
That space and new line.

src/nanovg.c
memononen: No need for extra functions, use variables :)
```C
float halfw = nvg__absf(w)*0.5f;
float halfh = nvg__absf(h)*0.5f;
...
```
Also, instead of 1,2,3, use radXTopLeft, or rxTL, etc.

src/nanovg.c
xeekworx: I agree with the name, the Ex kinda brought back bad win32api dreams.

The ordering of the radii parameters was because of the order they were being drawn, in my actual C++ wrapper it's just the way you suggested. It definitely makes sense.

src/nanovg.c
xeekworx: I don't recall changing the formatting, maybe Visual Studio did that to me automatically.

xeekworx:
New commit based on all of @memononen's suggestions.

src/nanovg.h
```diff
@@ -491,7 +491,7 @@ void nvgRoundedRect(NVGcontext* ctx, float x, float y, float w, float h, float r
 
 // Creates new rounded rectangle shaped sub-path with more radius control for each corner.
 // r1 is bottom left, r2 is bottom right, r3, is top right, r4 is top left.
```
memononen: One more thing, the comment. I think we can remove the second line.

xeekworx: Oh rats .. that one got by me!

xeekworx:
Things got pretty strange when I tried to amend my last commit, but it should look right after I merged in the conflict.

memononen:
Merged, thank you for the improvement!

PR #330, opened 2016-10-19
Fixes line breaks for CJK characters.

olliwang:
This commit fixes weird line breaks when rendering CJK (Chinese, Japanese, Korean) characters. Each CJK character is more like a word instead of a letter. This commit fixes the logic of line breaks by introducing a new `NVG_CJK_CHAR` code point type and takes special care for formatting text including CJK characters.

Before:
img_7226

After:
img_7227


memononen:
Thanks for the fix! These are the kind of details that are easy to miss when not using a certain language.

olliwang:
Thanks for merge. 👍

PR #331, opened 2016-10-20
Improves to display CJK characters with English words.

olliwang:
This commit improves to display CJK characters mixed with English words so there is less spaces at line breaks. It seems that the previous commit didn't handle CJK characters very well. This commit should fix the issue.

Before:
img_7246

After:
img_7245


memononen:
Thanks!

PR #333, opened 2016-10-22
Fixes the bug that text may not render correctly.

olliwang:
This commit fixes the bug that text may not render correctly when glyph is not found in fallback fonts. This bug is hard to find when rendering western letters because almost every font includes all western letters. But when it comes to Chinese characters, it's easily that a glyph is not found in western fonts and `fons__getGlyph()` should return `NULL` until meeting a Chinese font.

Before:
img_7267

After:
img_7269

### DIFF
--- a/src/fontstash.h
+++ b/src/fontstash.h
@@ -1039,6 +1039,7 @@ static FONSglyph* fons__getGlyph(FONScontext* stash, FONSfont* font, unsigned in
 	int pad, added;
 	unsigned char* bdst;
 	unsigned char* dst;
+	FONSfont* renderFont = font;
 
 	if (isize < 2) return NULL;
 	if (iblur > 20) iblur = 20;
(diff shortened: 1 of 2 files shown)


memononen:
This case also caught up my attention when I did the initial code review, and I came up with the conclusion that not returning `NULL` is correct, but looks like I was wrong :) I think the function is currently too subtle so that all the cases would be visible.

Firstly, the function should not only return `NULL` in case of error, i.e. not enough memory. In case it does not find a glyph, it should return empty glyph, and cache that result.

Secondly, I don't like how the function calls itself.

Maybe something like this could work:
```C
	FONSfont* renderFont = font;

	...

	// Try to find the glyph in fallback fonts.
	if (g == 0) {
		for (i = 0; i < font->nfallbacks; ++i) {
			FONSfont* fallbackFont = stash->fonts[font->fallbacks[i]];
			int fallbackIndex = fons__tt_getGlyphIndex(&fallbackFont->font, codepoint);
			if (fallbackIndex != 0) {
				g = fallbackIndex;
				renderFont = fallbackFont;
				break;
			}
		}
		// It is possible that we did not find a fallback glyph.
		// In that case the glyph index 'g' is 0, and we'll proceed below and cache empty glyph.
	}

	...
	scale = fons__tt_getPixelHeightScale(&renderFont->font, size);
	fons__tt_buildGlyphBitmap(&renderFont->font, g, size, scale, &advance, &lsb, &x0, &y0, &x1, &y1);

	...
	fons__tt_renderGlyphBitmap(&renderFont->font, dst, gw-pad*2,gh-pad*2, stash->params.width, scale,scale, g);
```

That way the miss gets cached earlier too.

olliwang:
@memononen Your solution indeed is much better. 👍  I've tested your code and it works well in my app. I also amended the commit, please let me know if something needs to modify.

src/fontstash.h
```diff
@@ -1057,18 +1059,21 @@ static FONSglyph* fons__getGlyph(FONScontext* stash, FONSfont* font, unsigned in
...
 	// Could not find glyph, create it.
-	scale = fons__tt_getPixelHeightScale(&font->font, size);
 	g = fons__tt_getGlyphIndex(&font->font, codepoint);
 	// Try to find the glyph in fallback fonts.
 	if (g == 0) {
 		for (i = 0; i < font->nfallbacks; ++i) {
-			FONSglyph* fallbackGlyph = fons__getGlyph(stash, stash->fonts[font->fallbacks[i]], codepoint, isize, iblur);
-			if (fallbackGlyph != NULL && fallbackGlyph->index != 0) {
-				return fallbackGlyph;
+			fallbackFont = stash->fonts[font->fallbacks[i]];
```
memononen: It is valid C to define these two variables (fallbackFont, fallbackIndex)  at the beginning of the scope here.

src/fontstash.h
```diff
@@ -1057,18 +1059,21 @@ static FONSglyph* fons__getGlyph(FONScontext* stash, FONSfont* font, unsigned in
...
+			fallbackIndex = fons__tt_getGlyphIndex(&fallbackFont->font, codepoint);
+			if (fallbackIndex != 0) {
+				g = fallbackIndex;
+				renderFont = fallbackFont;
+				break;
 			}
 		}
 	}
```
memononen: Let's add the comment about caching missing glyphs here too here too.

memononen:
Can you also check that the caching works in case of fallback font? When the cache is empty, it should hit the fallback codepath, after that it should always be in the cache.

Can you also test that with some non-existing glyph (i.e. some emoji with a font w/o emojis)?

Does the emoji font on the demo show properly?
https://github.com/memononen/nanovg/blob/master/example/demo.c#L863

olliwang:
Just updated the code and finished the tests. Caching works in case of fallback font. When hitting non-existing glyph, square is rendered as expected, and the emoji font on the demo shows without any problem. ;)

memononen:
Great! Thanks for the fix and testing.

PR #338, opened 2016-11-01
Better support for CJK characters.

olliwang:
Previously, the Unicode range of the CJK chractaters only includes Chinese charactaters, Japanese Kanji, and CJK Symbols and Punctuation. The range are used by traditioanl Korean as well. This commit further includes [Hiragana](https://en.wikipedia.org/wiki/Hiragana_(Unicode_block)) and [Katakana](https://en.wikipedia.org/wiki/Katakana_(Unicode_block)) in Japanese, and [Hangul Syllables](https://en.wikipedia.org/wiki/Hangul_Syllables) and [Hangul Jamo](https://en.wikipedia.org/wiki/Hangul_Jamo_(Unicode_block)) in Korean to make CJK support more complete.

Here's a Japanese Example.

Before:
img_7443

After:
img_7444

### DIFF
--- a/src/nanovg.c
+++ b/src/nanovg.c
@@ -2617,7 +2617,12 @@ int nvgTextBreakLines(NVGcontext* ctx, const char* string, const char* end, floa
 				type = NVG_NEWLINE;
 				break;
 			default:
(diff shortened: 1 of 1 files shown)


memononen:
Thanks for the upgrade!

PR #357, opened 2017-02-23
Bug fixes to compile under OS X 12.3

ckaran:
I've discovered two bugs which prevented NanoVG from compiling under OS X 12.3 correctly.  First, the Carbon framework was missing, which prevented correct linking as they are needed by GLFW.  Second, I discovered that `nanovg_gl.h` was missing the includes for `<OpenGL/glext.h>`.  I've made both fixes, and put them as separate commits in this pull request.

### DIFF
--- a/premake4.lua
+++ b/premake4.lua
@@ -5,7 +5,7 @@ solution "nanovg"
 	location ( "build" )
 	configurations { "Debug", "Release" }
 	platforms {"native", "x64", "x32"}
-	
+
    	project "nanovg"
 		language "C"
 		kind "StaticLib"
@@ -13,7 +13,7 @@ solution "nanovg"
 		files { "src/*.c" }
 		targetdir("build")
 		defines { "_CRT_SECURE_NO_WARNINGS" } --,"FONS_USE_FREETYPE" } Uncomment to compile with FreeType support
-		
+
 		configuration "Debug"
 			defines { "DEBUG" }
 			flags { "Symbols", "ExtraWarnings"}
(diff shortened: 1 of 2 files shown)


memononen:
Thank for the fixes!

ckaran:
No problem! :)  I'll close this out.

olliwang:
This commit is very bad. Including `<OpenGL/glext.h>` only works on Mac, but not on iOS. This commit causes compiling error for iOS app. In iOS, the header could be either `<OpenGLES/ES2/gl.h>` or `<OpenGLES/ES3/gl.h>`. So the header should be included manually based on the needs.

memononen:
@olliwang Thanks for reporting this, my bad. I think my thinking has been not to include any OpenGL headers in the GL backend because it's such a mess on different platforms. 

I think the correct way to fix this would be to change the `example_xxx.c` instead, and remove the header include from the GL backend:
```C
#include <stdio.h>
#define GLFW_INCLUDE_ES3
#define GLFW_INCLUDE_GLEXT
#include <GLFW/glfw3.h>
#include "nanovg.h"
#define NANOVG_GLES3_IMPLEMENTATION
#include "nanovg_gl.h"
#include "nanovg_gl_utils.h"
#include "demo.h"
#include "perf.h"
```
http://www.glfw.org/docs/latest/context_guide.html

I blame this on my mushy brain caused by taking care of our 5 month old baby ;)

@ckaran would you mind making a PR for the fix?

ckaran:
Is this fixed at this point?  Do you want me to take a stab at it?

memononen:
@olliwang 's #359 fixes it. Can you test if it works for you too?

ckaran:
@olliwang @memononen Just tested it, it looks like it all works for me!  Go ahead and close this out, if there are problems in the future, I'll open a new issue.

PR #359, opened 2017-03-06
Fixes how glext header is included.

olliwang:
This commit fixes the issue caused by #357 that nanovg failed to compile on iOS due to the wrong way of including glext header.


memononen:
Thanks!

olliwang:
Thanks for the quick merge. :D

PR #385, opened 2017-07-01
Discards fragments as early as possible.

olliwang:
This commit moves the `discard` instruction in the fragment shader to an earlier position to avoid performing calculations whose results are unused.

In my real world app, tens of thousands polygons may render per frame, and this simple change improves the FPS from 37 to 45 in a test case. It's not bad.

Before:
screen shot 2017-07-02 at 01 04 31

After:
screen shot 2017-07-02 at 01 05 57


memononen:
Good improvement, thanks!

PR #386, opened 2017-07-04
Allows to turn on/off antialias for shapes.

olliwang:
This commit adds `nvgEnable()` and `nvgDisable()` functions for changing flags after creating the `NVGcontext` object.

Allowing changing flags is useful to get some performance gain at certain situations. For example, when applying `nvgFill()` and `nvgStroke()`  to the same path with an opaque stroke color. The antialiasing for fill operation is unnecessary because the shape edge will be covered by stroke anyway. That way, I can do something like the following code to reduce some triangle vertices from processing.

    nvgDisable(context, NVG_ANTIALIAS);
    nvgFill(context);
    nvgEnable(context, NVG_ANTIALIAS);
    nvgStroke(context);

In my real world app. I need to stroke and fill thousands of polygons per frame, and this would help improve the FPS and battery consumption a bit. Here's a simple benchmark in a test case of my app running on iPhone 6s+.

|  | Before | After|
| --- | --- | --- |
| nverts | 169,628 | 137,340 |
| FPS | 38~39 | 44~45 |

### DIFF
--- a/src/nanovg.c
+++ b/src/nanovg.c
@@ -67,6 +67,7 @@ enum NVGpointFlags
 
 struct NVGstate {
 	NVGcompositeOperationState compositeOperation;
+	int shapeAntiAlias;
 	NVGpaint fill;
 	NVGpaint stroke;
 	float strokeWidth;
@@ -646,6 +647,7 @@ void nvgReset(NVGcontext* ctx)
 	nvg__setPaintColor(&state->fill, nvgRGBA(255,255,255,255));
 	nvg__setPaintColor(&state->stroke, nvgRGBA(0,0,0,255));
 	state->compositeOperation = nvg__compositeOperationState(NVG_SOURCE_OVER);
+	state->shapeAntiAlias = 1;
 	state->strokeWidth = 1.0f;
 	state->miterLimit = 10.0f;
 	state->lineCap = NVG_BUTT;
@@ -665,6 +667,12 @@ void nvgReset(NVGcontext* ctx)
 }
 
 // State setting
+void nvgShapeAntiAlias(NVGcontext* ctx, int enabled)
+{
+	NVGstate* state = nvg__getState(ctx);
+	state->shapeAntiAlias = enabled;
+}
+
 void nvgStrokeWidth(NVGcontext* ctx, float width)
 {
 	NVGstate* state = nvg__getState(ctx);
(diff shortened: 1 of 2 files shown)


olliwang:
@memononen Do you think there is anything needed to change to make this PR approved? Thanks.

memononen:
Sorry for not commenting on this earlier, I have been really busy.

I like the idea, and I think it is worth implementing. Based on your usage pattern, I think the flag should be part of `NVGstate`, though. So that you can save/restore state. That way it would be more like "drawing style", rather than backend capability.

With your implementation the shader does not change, so the flag change is not comparable to changing the creating flags. The speedup comes from not creating and rendering the extra geometry.

I suggest to change checks for `ctx->params.edgeAntiAlias` to `ctx->params.edgeAntiAlias && state->edgeAntiAlias`, enable/disable should change the state instead. Maybe call the flag `NVG_ANTIALIAS_HINT` (and make it the same value as `NVG_ANTIALIAS_HINT`, since it can only turn off antialias, not turn it on if the backend was created w/o AA.

The reason for calling it hint is that it will not work for text and maybe not strokes either. Well, device creating will not disable text AA either, still :) An alternative would be to call it something like "quick render" or something, which indicates that some corners are cut.

This change should also enable to implement a version of `drawImage` which renders just two triangles.

olliwang:
Thanks @memononen. I've rewritten the commit to add a new `nvgHints()` function for setting the `NVG_ANTIALIAS_HINT` flag. Please let me know if something needs to change.

I'm not sure what do you mean about the `drawImage` part. If I understand correctly, when filling an image by the combination of `nvgRect()`, `nvgFillPaint()`, and `nvgFill()` with `edgeAntiAlias` disabled. The `glnvg__convexFill()` should kick in and use `GL_TRIANGLE_FAN` to render the quad. In my test I do see the number of `paths[i].fillCount` is equal to 4.

memononen:
You did the thing I asked, but I think it does not feel quote right yet. The hint thing feels a bit premature future proofing, my bad.

This is specific optimization for the specific algorithm used by NanoVG, common to all backends
How about just calling it `nvgShapeAntialias(ctx, int enabled)`? And explain that it turns on/off antialiasing for `nvgFill` and `nvgStroke`.

There's been previous issue about rendering just sprites/images quicky. This change enables that case too, just like you describe.

olliwang:
@memononen Got it. Just updated. :)

memononen:
One more thing :) I think the shapeAA check is missing from `nvgStroke`.

olliwang:
You're right. My bad. Just updated. 👍

memononen:
Merged, thanks for the improvement :)

PR #387, opened 2017-07-11
Prevents uploading unused vertices to GPU.

olliwang:
This commit fixes the bug that `glnvg__renderFill()` allocates unused quad vertices for `GLNVG_CONVEXFILL`, which causes unused vertices uploaded to the GPU.

Besides, the quad for the `GLNVG_FILL` type is changed to drawn by `GL_TRIANGLE_STRIP` instead of `GL_TRIANGLES` so two more vertices can be reduced per quad.

In one of my test case, I need to paste an image to a framebuffer about 5,000 times, and this change would reduce the total number of uploaded vertices from `52,106` to `22,476`. It is good for FPS and battery consumption.

### DIFF
--- a/src/nanovg_gl.h
+++ b/src/nanovg_gl.h
@@ -1028,7 +1028,7 @@ static void glnvg__fill(GLNVGcontext* gl, GLNVGcall* call)
 	// Draw fill
 	glnvg__stencilFunc(gl, GL_NOTEQUAL, 0x0, 0xff);
 	glStencilOp(GL_ZERO, GL_ZERO, GL_ZERO);
-	glDrawArrays(GL_TRIANGLES, call->triangleOffset, call->triangleCount);
+	glDrawArrays(GL_TRIANGLE_STRIP, call->triangleOffset, call->triangleCount);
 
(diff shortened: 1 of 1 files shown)


src/nanovg_gl.h
```diff
@@ -1354,9 +1354,11 @@ static void glnvg__renderFill(void* uptr, NVGpaint* paint, NVGcompositeOperation
 
 	if (npaths == 1 && paths[0].convex)
 		call->type = GLNVG_CONVEXFILL;
+ 	else
```
memononen: This code is a bit hard to follow not. How about:
```C
...
call->type = GLNVG_FILL;
call->triangleCount = 4;
...
if (npaths == 1 && paths[0].convex) {
 	call->type = GLNVG_CONVEXFILL;
 	call->triangleCount = 0;  // Bounding box fill quad not needed for convex fill
}
```

olliwang:
@memononen Thanks for your advise. Just changed it. :)

memononen:
Thanks for the optimization!

olliwang:
You are welcome. 👍

PR #392, opened 2017-07-27
Fix example crash on 32bit platform

### LINKED ISSUE #377: example crash in stb_truetype.h
I run on armhf32 linux, ubuntu 14.04
In seven examples, only example_fbo can run correctly, other six examples all crash for same error:
example_gles3: ../src/stb_truetype.h:1700: stbtt__new_active: Assertion `z != ((void *)0)' failed.

>> #elif STBTT_RASTERIZER_VERSION == 2
>> static stbtt__active_edge *stbtt__new_active(stbtt__hheap *hh, stbtt__edge *e, int off_x, float start_point, void *userdata)
>> {
>>    stbtt__active_edge *z = (stbtt__active_edge *) stbtt__hheap_alloc(hh, sizeof(*z), userdata);
>>    float dxdy = (e->x1 - e->x0) / (e->y1 - e->y0);
>>    STBTT_assert(z != NULL);    ==> this line 1700

I found out, it's because example_fbo has a loadFonts function, and other six examples don't.
I copy the loadFonts function to other examples source and insert calling at :

	if (vg == NULL) {
		printf("Could not init nanovg.\n");
		return -1;
	}
	if (loadFonts(vg) == -1) {
		printf("Could not load fonts\n");
		return -1;
	}
	if (loadDemoData(vg, &data) == -1)
		return -1;
 then it fix the problem.

satoren:
Fix #377 

32bit platform needs  more buffer size then 64bit platform

https://github.com/memononen/nanovg/blob/master/src/stb_truetype.h#L1617

on 32bit 
`sizeof(stbtt__active_edge) == 28`
count is 2000

on 64bit 
`sizeof(stbtt__active_edge) == 32`
count is 800

### DIFF
--- a/src/fontstash.h
+++ b/src/fontstash.h
@@ -306,7 +306,7 @@ int fons__tt_getGlyphKernAdvance(FONSttFontImpl *font, int glyph1, int glyph2)
 #endif
 
 #ifndef FONS_SCRATCH_BUF_SIZE
-#	define FONS_SCRATCH_BUF_SIZE 64000
+#	define FONS_SCRATCH_BUF_SIZE 96000
 #endif
 #ifndef FONS_HASH_LUT_SIZE
 #	define FONS_HASH_LUT_SIZE 256


memononen:
Thanks for the fix!

PR #395, opened 2017-07-31
Adds link of the Metal port.

olliwang:
I just implmeneted the native Metal port of NanoVG. ;)
https://github.com/ollix/MetalNanoVG


memononen:
Nice! What is the performance compared to OGL? Not sure how interpret the precentages in the table. 75% faster?

olliwang:
@memononen The percentages are the CPU usage reported by Xcode. If an iOS device has two cores, then the total will be 200%. When running the demo app, Metal uses 20% while OpenGL uses around 35%.

memononen:
@olliwang Ok, so I did read it correctly :) Good improvement.

PR #405, opened 2017-09-05
Fix MSVC 2017 compiler warning

Joshua-Ashton:
This fixes a compiler warning when using Visual Studio 2017 to compile. 
It warns about the conversion from int to float on the line with the ternary operator determining the textype. 
Textype can be an int or float depending on whether NANOVG_GL_USE_UNIFORMBUFFER is defined.
This takes that into account using the aforementioned define to avoid this warning.


memononen:
Thanks for the fix!

PR #406, opened 2017-09-05
Use external libraries as git submodules instead of redisting.

Joshua-Ashton:
Removed the redistributing of stb and fontstash dependencies.
Uses git submodules instead.


Joshua-Ashton:
To whoever thumbed down, please explain why or add to discussion 🙁

fundamental:
I just saw the comment, you can hover over the 'reaction' to see who responded in the future and @ them if you want to refer to them directly. To elaborate, git submodules add complexity in order to manage complex sub-project dependencies. In the case of stb/fontstash/etc those projects are designed to have a minimal footprint such that it's easy to copy/paste a few files and be done with it. Git submodules add more complexity to working with the nanovg repo, but in my opinion don't justify themselves in the complexity that they remove (managing some stb/fontstash changes).

I do use submodules elsewhere in projects I maintain, but that's to manage sub-projects with 50+ files, not 1-2.

Joshua-Ashton:
You literally just add --recursive to your git pull. That's not complex.

PR #411, opened 2017-09-22
Support ~2010 Mac graphics cards (and possibly others) with nvgluCreateFramebuffer patch

### LINKED ISSUE #410: example_fbo cannot run on certain Mac computers
I suspect from [this issue](https://github.com/VCVRack/Rack/issues/17) in the VCV Rack repo that the Mac computers supporting up to OpenGL 3.3 on [this list](https://support.apple.com/en-us/HT202823) cannot create framebuffers using `nvgluCreateFramebuffer` in nanovg_gl_util.h. I suspect this is because the graphic cards do not support stencil renderbuffers, but I am not 100% sure.

Is there a way around this? I am not an expert in advanced OpenGL features, so I would be grateful to have any insight into alternatives for supporting these Macs while supporting the feature nanovg requires, since I heavily use framebuffers in my application.

`glCheckFramebufferStatus(GL_FRAMEBUFFER)` returns `GL_FRAMEBUFFER_UNSUPPORTED` inside that function.

AndrewBelt:
Fixes #410

### DIFF
--- a/src/nanovg_gl_utils.h
+++ b/src/nanovg_gl_utils.h
@@ -90,7 +90,15 @@ NVGLUframebuffer* nvgluCreateFramebuffer(NVGcontext* ctx, int w, int h, int imag
 	glFramebufferTexture2D(GL_FRAMEBUFFER, GL_COLOR_ATTACHMENT0, GL_TEXTURE_2D, fb->texture, 0);
 	glFramebufferRenderbuffer(GL_FRAMEBUFFER, GL_STENCIL_ATTACHMENT, GL_RENDERBUFFER, fb->rbo);
 
-	if (glCheckFramebufferStatus(GL_FRAMEBUFFER) != GL_FRAMEBUFFER_COMPLETE) goto error;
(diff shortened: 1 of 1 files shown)


memononen:
Thanks for the fix!

gavinhoward:
This commit breaks compilation on Linux for example_gles2. It says that the symbol `GL_DEPTH24_STENCIL8` is undeclared.

memononen:
@gavinhoward Merged your PR, thanks! `GL_DEPTH24_STENCIL8` indeed is not part of GLES2.

Is `GL_OES_packed_depth_stencil` supported on your platform?
https://www.khronos.org/registry/OpenGL/extensions/OES/OES_packed_depth_stencil.txt

If it is, can you check if it works? I'm pretty sure the depth/stencil is only required on some nvidia or radeon cards, but not generally on gles2 devices.

gavinhoward:
I will check tomorrow; I don't program on Sundays as a rule (sorry!).

gavinhoward:
Oh, and I should clarify that @AndrewBelt is the one who made this pull request. I am a bit too new to OpenGL to know how to fix it.

gavinhoward:
My problem has been fixed. Thank you!

PR #412, opened 2017-09-23
Only apply fallback if GL_DEPTH24_STENCIL8 is defined

AndrewBelt:
Fixes https://github.com/memononen/nanovg/pull/411#issuecomment-331638190

All examples compile and run.


memononen:
Thanks for the fix!

gavinhoward:
This fixes my problem as well!

PR #419, opened 2017-10-04
Make it easier to build macOS examples

wtholliday:
Make it easier to build macOS examples. Include build instructions in README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,16 @@ The data for the whole frame is buffered and flushed in `nvgEndFrame()`. The fol
 	glBindTexture(GL_TEXTURE_2D, tex);
 	glUniformBlockBinding(... , GLNVG_FRAG_BINDING);
 ```
+## Building the Examples
+
+### macOS
+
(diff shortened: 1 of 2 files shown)


memononen:
I'm not using brew myself, so I will not advocate using it because I cannot
do support for it. I have downloaded the library from glfw.org and build it
from the package. I think it builds static library and as per the developer
site, you should use glfw3 in that case:
http://www.glfw.org/docs/latest/build_guide.html


--mikko

wtholliday:
AFAIK, most people use homebrew these days.

wtholliday:
Note #380 is asking for the same thing.

lieff:
Still I think glfw || glfw3 must be auto-detected by build scripts. If we stick to one variant we loose one of this: Travis CI autobuild, native glfw build, brew, macports. But Travis CI for example is very useful.

wtholliday:
@lieff That sounds good! I think open source projects are crippled when there is friction in getting them to build, and so many people overlook that. Just trying to help.

PR #430, opened 2017-10-29
[Discussion - DO NOT MERGE] Initial work on multi-stop gradients.

Tribex:
I'm working on adding support for multi-stop gradients in NanoVG.

This PR contains a bit of a mashup of the current API and a potential new API design that uses the `image` field in `NVGpaint` for holding the gradient data. It also modifies the fragment shader to interpolate any number of stops from the gradient image.

At the moment, only the buttons in the example use this new method. Everything else uses the old method. If this were to be implemented in master the two methods would need to be properly separated out.

So here's what I've got so far and what I'd love help with:
1. Right now, gradients are stored in an RGBA texture, where each color is a single pixel expanding to the right. With GL3, this produces fairly accurate gradients. However, in GL2 as far as I know, it's impossible to get the texture width from the shader without passing it in, so a much more rudimentary method is used that isn't as accurate. You can see this difference in example_gl2 vs example_gl3. To resolve this we'd either need to pass in the texture width or find a better method that doesn't rely on the width.
2. The gradient on background color blending method seems different between the current gradient implementation vs. the one in this patch. I'm not sure what exactly the issue is.
```c
// demo.c:822 - 824
// Produces roughly equivalent results to the current method in the new method:
	buttonGradientSource[0] = nvgRGBA(50, 50, 50, 32);
	buttonGradientSource[1] = nvgRGBA(0, 0, 0, 32);
// Colors used in the current method:
	buttonGradientSource[0] = nvgRGBA(255, 255, 255, 32);
	buttonGradientSource[1] = nvgRGBA(0, 0, 0, 32);
```
3. Right now all color stops are equidistant. To support actual varying stop distances we'd probably need to pass in another texture. (Not sure what format would be best.) In which case, the gradient might itself be moved into a separate texture from the image. I'm not experienced enough with OpenGL to do that at the moment, as I think it would require some big internal changes.
4. I'm not a fan of the int -> float -> int conversion of colors. It might be worth using a texture format with float colors, but again, I'm not experienced enough to do that yet.

Thanks! I'm really hoping that something like this might end up in NanoVG someday, as lack of multi-stop gradients is one of the main blockers for using NanoVG in a project at the moment.

### DIFF
--- a/example/demo.c
+++ b/example/demo.c
@@ -87,7 +87,7 @@ void drawWindow(NVGcontext* vg, const char* title, float x, float y, float w, fl
 	nvgFill(vg);
 
 	// Header
-	headerPaint = nvgLinearGradient(vg, x,y,x,y+15, nvgRGBA(255,255,255,8), nvgRGBA(0,0,0,16));
+	headerPaint = nvgLinearGradient(vg, x,y,x,y+15, nvgRGBA(255,255,255,8), nvgRGBA(0,0,0,16), 0);
 	nvgBeginPath(vg);
 	nvgRoundedRect(vg, x+1,y+1, w-2,30, cornerRadius-1);
 	nvgFillPaint(vg, headerPaint);
@@ -157,7 +157,7 @@ void drawDropDown(NVGcontext* vg, const char* text, float x, float y, float w, f
 	char icon[8];
 	float cornerRadius = 4.0f;
 
-	bg = nvgLinearGradient(vg, x,y,x,y+h, nvgRGBA(255,255,255,16), nvgRGBA(0,0,0,16));
+	bg = nvgLinearGradient(vg, x,y,x,y+h, nvgRGBA(255,255,255,16), nvgRGBA(0,0,0,16), 0);
 	nvgBeginPath(vg);
 	nvgRoundedRect(vg, x+1,y+1, w-2,h-2, cornerRadius-1);
 	nvgFillPaint(vg, bg);
@@ -269,14 +269,14 @@ void drawCheckBox(NVGcontext* vg, const char* text, float x, float y, float w, f
 	nvgText(vg, x+9+2, y+h*0.5f, cpToUTF8(ICON_CHECK,icon), NULL);
 }
 
-void drawButton(NVGcontext* vg, int preicon, const char* text, float x, float y, float w, float h, NVGcolor col)
+void drawButton(NVGcontext* vg, DemoData* data, int preicon, const char* text, float x, float y, float w, float h, NVGcolor col)
 {
 	NVGpaint bg;
 	char icon[8];
 	float cornerRadius = 4.0f;
 	float tw = 0, iw = 0;
 
-	bg = nvgLinearGradient(vg, x,y,x,y+h, nvgRGBA(255,255,255,isBlack(col)?16:32), nvgRGBA(0,0,0,isBlack(col)?16:32));
+	bg = nvgLinearGradient(vg, x,y,x,y+h, col, col, isBlack(col) ? data->gradients.buttonBlack : data->gradients.buttonColored);
 	nvgBeginPath(vg);
 	nvgRoundedRect(vg, x+1,y+1, w-2,h-2, cornerRadius-1);
 	if (!isBlack(col)) {
@@ -335,7 +335,7 @@ void drawSlider(NVGcontext* vg, float pos, float x, float y, float w, float h)
 	nvgFill(vg);
 
 	// Knob Shadow
-	bg = nvgRadialGradient(vg, x+(int)(pos*w),cy+1, kr-3,kr+3, nvgRGBA(0,0,0,64), nvgRGBA(0,0,0,0));
+	bg = nvgRadialGradient(vg, x+(int)(pos*w),cy+1, kr-3,kr+3, nvgRGBA(0,0,0,64), nvgRGBA(0,0,0,0), 0);
 	nvgBeginPath(vg);
 	nvgRect(vg, x+(int)(pos*w)-kr-5,cy-kr-5,kr*2+5+5,kr*2+5+5+3);
 	nvgCircle(vg, x+(int)(pos*w),cy, kr);
@@ -344,7 +344,7 @@ void drawSlider(NVGcontext* vg, float pos, float x, float y, float w, float h)
 	nvgFill(vg);
 
 	// Knob
-	knob = nvgLinearGradient(vg, x,cy-kr,x,cy+kr, nvgRGBA(255,255,255,16), nvgRGBA(0,0,0,16));
+	knob = nvgLinearGradient(vg, x,cy-kr,x,cy+kr, nvgRGBA(255,255,255,16), nvgRGBA(0,0,0,16), 0);
 	nvgBeginPath(vg);
 	nvgCircle(vg, x+(int)(pos*w),cy, kr-1);
 	nvgFillColor(vg, nvgRGBA(40,43,48,255));
@@ -373,14 +373,14 @@ void drawEyes(NVGcontext* vg, float x, float y, float w, float h, float mx, floa
 	float br = (ex < ey ? ex : ey) * 0.5f;
 	float blink = 1 - pow(sinf(t*0.5f),200)*0.8f;
 
-	bg = nvgLinearGradient(vg, x,y+h*0.5f,x+w*0.1f,y+h, nvgRGBA(0,0,0,32), nvgRGBA(0,0,0,16));
+	bg = nvgLinearGradient(vg, x,y+h*0.5f,x+w*0.1f,y+h, nvgRGBA(0,0,0,32), nvgRGBA(0,0,0,16), 0);
 	nvgBeginPath(vg);
 	nvgEllipse(vg, lx+3.0f,ly+16.0f, ex,ey);
 	nvgEllipse(vg, rx+3.0f,ry+16.0f, ex,ey);
 	nvgFillPaint(vg, bg);
 	nvgFill(vg);
 
-	bg = nvgLinearGradient(vg, x,y+h*0.25f,x+w*0.1f,y+h, nvgRGBA(220,220,220,255), nvgRGBA(128,128,128,255));
+	bg = nvgLinearGradient(vg, x,y+h*0.25f,x+w*0.1f,y+h, nvgRGBA(220,220,220,255), nvgRGBA(128,128,128,255), 0);
 	nvgBeginPath(vg);
 	nvgEllipse(vg, lx,ly, ex,ey);
 	nvgEllipse(vg, rx,ry, ex,ey);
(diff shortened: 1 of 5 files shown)


memononen:
I think the best way to implement gradients would be to have a gradient texture where all the gradients are stored. You currently create an image per gradient, which is really expensive. 

Each gradient should be i.e. 1x256 image. And example how to create such image is here: https://github.com/memononen/nanosvg/blob/master/src/nanosvgrast.h#L1258

Creating a gradient with multiple stops should behave similarly as current gradient creation (that is, no fiddling with image indices). We would need new API point for that, though. 

One option would be to have:
```C
NVGpaint nvgLinearGradientWithStops(NVGcontext* ctx, float sx, float sy, float ex, float ey, NVGcolor* colors, float* stops, int count);
```
alternative (note the `pattern` in the name):
```C
NVGpaint nvgLinearGradientPattern(NVGcontext* ctx, float sx, float sy, float ex, float ey);
void nvgAddColorStop(NVGpaint* paint, float position, NVGcolor colo);
```
(ditto for box and radial).  `nvgLinearGradient` & co would be marked as deprecated.

NVGpaint would contain a fixed number of color stops (i.e. 8).

lieff:
If you want make it compatible with Flash gradients, NVGpaint must contain at least 15 color stops. Also rotation must be supported, float sx, float sy, float ex, float ey is not enough. For radial gradients, settable focal point also needed.
Here my sample code for making linear and radial gradient textures as well: https://github.com/lieff/lvg/blob/master/render/common.c#L158

Tribex:
@memononen Ah, interesting idea. So is what you're saying something along the lines of having a global say, 256x256 texture for storing gradients, where each gradient is stored per-line (with a max of 256 gradients?) Or am I misinterpreting you? I'll take a look at the code later. Would each color be stored in a single pixel or would they be stored according to color stops?

The idea with the implementation here is to use (yes, lots of little textures) but have all the gradient calculation and interpolation done on the GPU instead of the CPU. My current idea for stops is to add a second line of "colors" to the gradient texture that is used for determining the position of the colors on the previous line. A bit wasteful, but should be possible. I'm honestly not sure what the optimal approach would be. Since gradients would only be created once, however, I suspect your idea would be faster, if less accurate. Compute the gradient texture on the CPU then upload it to the GPU to render directly.

With my current plan, an example texture might look like: 
```c
{
  0xFF0000FF, 0x0000FFFF, 0x00FF00FF,
  0x00000000, 0x0000002A, 0x000000FF
}
```
Which would be computed in the shader to mean something along the lines of ``red - blue -------- green```.

What would be the format you recommend batching gradients together with?

@lieff Oooh, another implementation. I'll take a look at that too.

memononen:
I prefer not to add new structs to the API.

In gradient texture the gradient is "flattened" into the texture so that gradient 0% is first pixel and 100% at last pixel. 256 is not super accurate, but not super memory intensive either.
 
Dealing with, say, 15 stops using just shader is not impractical, but I suspect it's slow because you need to read a lot of texture data, and either do a search or a big branch. 256x256 texture is too big IMHO (especially with float data), I'd start with 32x256, have the gradients persist for one frame, and grow the texture as needed.

Tribex:
@memononen Ah, makes sense.

I'm still having a bit of trouble figuring out what your intent is for the gradient texture though. Would all gradients for every draw call be stored on the same texture, or would there be a separate texture for each call?
My apologies, I'm not great at figuring out what people mean. (Or making myself clear.)

memononen:
Yes, one global texture where all the gradients go. It should also contain some kind of invalidation. So if a gradient is not used, we reuse the slot when a new gradient needs to ne allocated.

Tribex:
@memononen Alright. If I ever find time I'll try and get something like that implemented.

Tribex:
@memononen Alright, so I think I have the mechanism for creating, updating, and validating lines for the texture worked out and implemented. It all seems to be working fairly performantly.
At the moment though the code is a *bit* of a massive mess.
We'll probably have to split all of that gradient logic into a separate function.

Anyway, I'll have to do some digging before I can figure out when, where, and how the gradient texture should be bound. (Feel free to help out with that. OpenGL and I don't get along very well sometimes.)

Any thoughts? Is this a little crazy and overcomplicated?

Tribex:
Haven't made any progress yet. Still trying to hook up the sampler properly. This is confusing.

RodrigoSantiago:
I did a simple change in the shader, **GLNVGfragUniforms**, **NVGpaint** , some constants and the gradients work with up to 16 colors (more than enough) for linear, radial and box

Tribex:
@RodrigoSantiago Are the stops adjustable? If so, I'd love to see your codebase. :D

RodrigoSantiago:
These are simple changes, but they stay in separate places in the code

```
struct NVGpaint {
        [...]
	int image; // last line
	int multStopEnabled;
	int stopsCount;
	int cycleMethod;
	int interpolation;
	float stops[16];
	float colors[16];
};
```
```
    struct GLNVGfragUniforms {
     [...] 
	int type; // last line
        int multStopEnabled;
        int stopsCount;
        int cycleMethod;
        int interpolation;
        float stops[16];
        float colors[16];
    [...]
	#define NANOVG_GL_UNIFORMARRAY_SIZE 20
     [...] 
	int type; // last line
        int multStopEnabled;
        int stopsCount;
        int cycleMethod;
        int interpolation;
        float stops[16];
        float colors[16];
```

```
        static int glnvg__convertPaint [...]
            frag->multStopEnabled = paint->multStopEnabled;
            frag->stopsCount = paint->stopsCount;
            frag->cycleMethod = paint->cycleMethod;
            frag->interpolation = paint->interpolation;

            for (int i = 0; i < 16; i++) {
                frag->stops[i] = paint->stops[i];
                frag->colors[i] = glnvg__premulColorF(paint->colors[i]);
            }
```

```
static float glnvg__premulColorF(float color) {
    typedef union {
        float f;
        int   i;
    } FloatBits;

    static FloatBits fb;

    fb.f = color;
    float a = (fb.i & 0xFF) / 255.0;
    int r = (int)(((fb.i >> 24) & 0xFF) * a);
    int g = (int)(((fb.i >> 16) & 0xFF) * a);
    int b = (int)(((fb.i >> 8) & 0xFF) * a);
    fb.i = r << 24 | g << 16 | b << 8 | (fb.i & 0xFF);
    return fb.f;
}
```

```
    SHADER : 
    [...]
    #define UNIFORMARRAY_SIZE 20
    [...]
    layout(std140) uniform frag {
    [...]
        int type; \\ last line
        int multStopEnabled;
        int stopsCount;
        int cycleMethod;
        int interpolation;
        vec4 stops[4];
        vec4 colors[4];
    [...]
    #ifndef USE_UNIFORMBUFFER
    [...]
	#define type int(frag[10].w) // last line
        #define multStopEnabled int(frag[11].x)
        #define stopsCount int(frag[11].y)
        #define cycleMethod int(frag[11].z)
        #define interpolation int(frag[11].w)
        #define stops vec4[](frag[12], frag[13], frag[14], frag[15])
        #define colors vec4[](frag[16], frag[17], frag[18], frag[19])

    [...] // added
    vec4 decodeColor(int index) {
       int rgba = floatBitsToInt(colors[index / 4][index % 4]);
            return vec4(
               ((rgba >> 24) & 0xFF) / 255.0, 
                ((rgba >> 16) & 0xFF) / 255.0, 
                ((rgba >>  8) & 0xFF) / 255.0, 
                ((rgba >>  0) & 0xFF) / 255.0);
    }
    float decodeStop(int index) {
        return stops[index / 4][index % 4];
    }
    vec4 decodeMix(vec4 a, vec4 b, float t) {\n"
        t = clamp(t, 0.0, 1.0);\n"
        if (interpolation == 1) t = sqrt(1 - (1 - t) * (1 - t));\n"
        else if (interpolation == 2) t = 1 - sqrt(1 - t * t);\n"
        else if (interpolation == 3) t = t * t * t * (t * (t * 6 - 15) + 10);\n"
        return mix(a, b, t);
    }
[...] // inside main
    if (type == 0) {			// Gradient
        // Calculate gradient color using box gradient
	vec2 pt = (paintMat * vec3(fpos,1.0)).xy;
       	float d = (sdroundrect(pt, extent, radius) + feather*0.5) / feather;
	vec4 color;
	// Multstop gradient
        if (multStopEnabled == 1) {
               if (cycleMethod == 1) d = d - floor(d);
               else if (cycleMethod == 2) d = (int(floor(d)) % 2 == 0) ? d - floor(d) : 1 - (d - floor(d));
               else d = clamp(d, 0.0, 1.0);
               int choose = stopsCount - 1;
               float stop = 0;
               float nStop = 0;
               for (int i = 0; i < stopsCount; i++) {
                   nStop = decodeStop(i);
                   if (d < nStop) {
                       choose = i - 1;
                       break;
                   }
                   stop = nStop;
               }
               color = decodeMix(decodeColor(choose), decodeColor(choose + 1), (d - stop) / (nStop - stop));
           } else {
	        color = decodeMix(innerCol, outerCol, clamp(d, 0.0, 1.0));
	    }
       	    // Combine alpha
	    color *= strokeAlpha * scissor;
	    result = color;
```
multStopEnabled = if is enabled
stopsCount = number of stops (up to 16)
cycleMethod =  (0 - clamp | 1 - reflect | 2 - repeate)
interpolation = (0 - linear | 1 - circleIn | 2 - circleOut | 3 - fade (i use for better shadows))

usage example :
```
NVGpaint p = nvgRadialGradient(vg, x, y, radiusIn, radiusOut, nvgRGBAf(0,0,0,1), nvgRGBAf(1, 1, 1, 1));
    p.multStopEnabled = 1;
    p.cycleMethod = 0;
    p.interpolation = 0;
    p.stopsCount = 3;
    p.stops[0] = 0.1;
    p.stops[1] = 0.5;
    p.stops[2] = 0.7;
    p.colors[0] = rgbaFloat(255, 0, 0, 1);
    p.colors[1] = rgbaFloat(0, 255, 0, 1);
    p.colors[2] = rgbaFloat(0, 0, 255, 1);
```

Dimension4:
Alright I figured it out myself. I've changed the gradient to behave like `innerColor -> stop0 -> stop1 -> stop2 -> outerColor`. Here is my version:
```glsl
if (multStopEnabled == 1)
{
    if (cycleMethod == 1) // reflect
        d = d - floor(d);
    else if (cycleMethod == 2) // repeat
        d = (int(floor(d)) % 2 == 0) ? d - floor(d) : 1 - (d - floor(d));
    else // clamp
        d = clamp(d, 0.0, 1.0);

    float prevStop;
    float stop = 0;
    int stopIdx = 0;

    for (; stopIdx < stopsCount; stopIdx++)
    {
        prevStop = stop;
        stop = decodeStop(stopIdx);
        if (d <= stop)
            break;
    }
    if (stopIdx == 0)
        color = decodeMix(innerCol, decodeColor(stopIdx), d / stop);
    else if (stopIdx == stopsCount)
        color = decodeMix(decodeColor(stopIdx - 1), outerCol, (d - stop) / (1 - stop));
    else
        color = decodeMix(decodeColor(stopIdx - 1), decodeColor(stopIdx), (d - prevStop) / (stop - prevStop));
}
else
{
    color = decodeMix(innerCol, outerCol, clamp(d, 0.0, 1.0));
}
```
About the encoding of the `stops` and `colors` variables. Is this a memory alignment thing? Or is it needed to be compliant in case `USE_UNIFORMBUFFER` is not defined?

PR #431, opened 2017-11-03
Resets font atlas when it is full.

olliwang:
When font atlas is full, the `fons__getGlyph()` function simply returns `NULL` and makes a lot of corresponded functions return unreliable results.

For example, when trying to auto scale text’s font size to fit a rectangle. The text may set a large font size initially and repeatedly calling `nvgTextBounds()` and reducing font size until the text can fit well.

This process may create a lot of one-time use font images and quickly makes the atlas full. As a result, `nvgTextBounds()` doesn't return a reliable result anymore because no new glyph can be generated.

This commit implements the error callback to reset the atlas when receiving the `FONS_ATLAS_FULL` error.

### DIFF
--- a/src/fontstash.h
+++ b/src/fontstash.h
@@ -38,6 +38,11 @@ enum FONSalign {
 	FONS_ALIGN_BASELINE	= 1<<6, // Default
 };
 
+enum FONSglyphBitmap {
+	FONS_GLYPH_BITMAP_OPTIONAL = 1,
+	FONS_GLYPH_BITMAP_REQUIRED = 2,
+};
+
 enum FONSerrorCode {
 	// Font atlas is full.
 	FONS_ATLAS_FULL = 1,
@@ -78,6 +83,7 @@ struct FONStextIter {
 	const char* next;
 	const char* end;
 	unsigned int utf8state;
+	int bitmapOption;
 };
 typedef struct FONStextIter FONStextIter;
 
@@ -122,7 +128,7 @@ void fonsLineBounds(FONScontext* s, float y, float* miny, float* maxy);
 void fonsVertMetrics(FONScontext* s, float* ascender, float* descender, float* lineh);
 
 // Text iterator
-int fonsTextIterInit(FONScontext* stash, FONStextIter* iter, float x, float y, const char* str, const char* end);
+int fonsTextIterInit(FONScontext* stash, FONStextIter* iter, float x, float y, const char* str, const char* end, int bitmapOption);
 int fonsTextIterNext(FONScontext* stash, FONStextIter* iter, struct FONSquad* quad);
 
 // Pull texture changes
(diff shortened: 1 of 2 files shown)


memononen:
The logic around `nvg__allocTextAtlas` should handle this. If that is not enough, one option would be to make the text dimension functions to not to rasterize the glyphs at all.

olliwang:
Thanks, @memononen. `nvg__allocTextAtlas()` is not enough for my case. I will try the option you suggested. There are a lot of functions depending on `fons__getGlyph()`. Should it accept a new parameter for whether to rasterize? It seems all `fons__getGlyph()` calls in `nvgTextBounds()` should pass rasterization to avoid atlas full.

memononen:
I'm thinking out loud, sorry if this is a bit inconsistent.

I think one option would be to add a flag to the `fons__getGlyph`, in which case we should not call `fons__atlasAddRect` (gx=gy=0) and `fons__tt_renderGlyphBitmap` and skip the blurring. So we would add an non-rendered glyph into the cache.

When the user requests the same glyph with rendering, we could either fall through without allocating new glyph, refilling the data (#1). Or we could let the empty glyph be there, and differentiate rasterized and empty glyphs, just like we check for different size and blur (#2). This effectively just would cache the data from the calls to the truetype library.

An alternative would be to have a second function(#3): 
```C
static int fons__getGlyphData(FONScontext* stash, FONSglyph* glyph, FONSfont* font, unsigned int codepoint, short isize, short iblur)
```
which would call true type API to the fill `glyph`.

I think I would prefer #1 of that does not end up too complicated, then #2.

Since some of the text functionality are done in NanoVG, you should also add something to `fonsTextIterInit` too.

olliwang:
Thank you. That's much clear. I'll try #1 and see what I can do. ;)

olliwang:
@memononen Just updated the code to implement #1. Tested it on my app and works like a charm. :)

src/fontstash.h
```diff
@@ -1380,7 +1410,7 @@ int fonsTextIterNext(FONScontext* stash, FONStextIter* iter, FONSquad* quad)
 		// Get glyph and quad
 		iter->x = iter->nextx;
 		iter->y = iter->nexty;
-		glyph = fons__getGlyph(stash, iter->font, iter->codepoint, iter->isize, iter->iblur);
+		glyph = fons__getGlyph(stash, iter->font, iter->codepoint, iter->isize, iter->iblur, FONS_GLYPH_BITMAP_REQUIRED);
```
memononen: There should be an iterator which calls getGlyph using optional. That way `nvgTextGlyphPositions` and `nvgTextBreakLines`, `nvgTextBoxBounds` will not render glyphs either. 

Maybe the best way would be to add additional parameter to `fonsTextIterInit`. The `fonsTextIterNext ` function declaration should have comment that if the iterator was initialized with optional, then the UV coordinates of the quad will be invalid.

olliwang:
@memononen Just updated. Now only `nvgText()` requires bitmap data. ;)

memononen:
`fons__tt_buildGlyphBitmap()` is not super expensive, so it's ok to call it again. This loioks good now, I'll merge it. Thanks for the improvement!

olliwang:
@memononen Thanks for your guide. The final result is much better than my initial solution, and it significantly improves the UI initialization time of my app. :D

memononen:
@olliwang excellent :)

PR #442, opened 2017-12-18
Mark devicePixelRatio unused in glnvg_renderViewport

gavinhoward:
This silences a warning in Clang (and GCC?) when -Wall and -Wextra is turned on.


memononen:
Thanks for the fix!

PR #452, opened 2018-01-22
COMP: as per memononen's suggestions, tweak findTexture

Chaircrusher:
This is based closely on the suggestion you gave to me, but tweaked to work a little better.

1. The salt is shifted once to the high word.  Right-shifting & left-shifting was messing up on ints.
2. I make the salt by caling rand; it appears that you were taking it from uninitialized memory, which isn't a reliable way to get unique salt for each texture.
3. I made sure the salt was nonzero, because if the salt was zero and the ID was zero, it makes findTexture look like it failed.

This solves the 'stale texture ID' problem:
1. If it's a valid ID, the salt matches.
2. If it's a deleted ID, the salt in the array is zero, and the salt in the ID doesn't match.
3. If it's a texture that's been deleted and re-used, the salt won't match.

This version has slightly more overhead than the original, (mostly on allocTexture when it matters least), but has better overall performance due to skipping a sequential search for textures by ID.

### DIFF
--- a/src/nanovg_gl.h
+++ b/src/nanovg_gl.h
@@ -147,6 +147,7 @@ struct GLNVGtexture {
 	int width, height;
 	int type;
 	int flags;
+	unsigned int salt;
 };
 typedef struct GLNVGtexture GLNVGtexture;
 
@@ -345,12 +346,12 @@ static void glnvg__blendFuncSeparate(GLNVGcontext* gl, const GLNVGblend* blend)
 
 static GLNVGtexture* glnvg__allocTexture(GLNVGcontext* gl)
 {
-	GLNVGtexture* tex = NULL;
-	int i;
+ 	GLNVGtexture* tex = NULL;
+	int i, idx = -1, salt;
 
 	for (i = 0; i < gl->ntextures; i++) {
 		if (gl->textures[i].id == 0) {
-			tex = &gl->textures[i];
+			idx = i;
 			break;
 		}
 	}
@@ -363,38 +364,54 @@ static GLNVGtexture* glnvg__allocTexture(GLNVGcontext* gl)
 			gl->textures = textures;
 			gl->ctextures = ctextures;
 		}
-		tex = &gl->textures[gl->ntextures++];
+		idx = gl->ntextures++;
 	}
 
+	tex = &gl->textures[idx];
+	while(!(salt = (unsigned int)rand()))
+		;
 	memset(tex, 0, sizeof(*tex));
-	tex->id = ++gl->textureId;
+	tex->salt = salt;
+	// salt value is in high 2 bytes, ID in low two bytes
+	tex->salt <<= 16;
+	tex->id = (idx & 0xffff) | tex->salt;
 
 	return tex;
 }
 
 static GLNVGtexture* glnvg__findTexture(GLNVGcontext* gl, int id)
 {
-	int i;
-	for (i = 0; i < gl->ntextures; i++)
-		if (gl->textures[i].id == id)
-			return &gl->textures[i];
-	return NULL;
+	int idx = id & 0xffff, salt = ((unsigned int)id) & 0xffff0000;
+	// Check that the texture index is valid
+	if (idx < 0 || idx >= gl->ntextures)
+		return NULL;
+	// Check that we're not trying to access deleted texture.
+	if (gl->textures[idx].salt != salt)
+		return NULL;
+	return &gl->textures[idx];
 }
 
 static int glnvg__deleteTexture(GLNVGcontext* gl, int id)
 {
-	int i;
-	for (i = 0; i < gl->ntextures; i++) {
-		if (gl->textures[i].id == id) {
-			if (gl->textures[i].tex != 0 && (gl->textures[i].flags & NVG_IMAGE_NODELETE) == 0)
-				glDeleteTextures(1, &gl->textures[i].tex);
-			memset(&gl->textures[i], 0, sizeof(gl->textures[i]));
-			return 1;
-		}
-	}
-	return 0;
-}
+	int idx = id & 0xffff, salt = ((unsigned int)id) & 0xffff0000;
+	GLNVGtexture* tex = NULL;
+	// Check that the texture index is valid
+	if (idx < 0 || idx >= gl->ntextures)
+		return 0;
+	// Check that we're not trying to access deleted texture.
+	if (gl->textures[idx].salt != salt)
+		return 0;
 
+	// Clear texture
+	tex = &gl->textures[idx];
+	if (tex->tex != 0 && (tex->flags & NVG_IMAGE_NODELETE) == 0)
+		glDeleteTextures(1, &tex->tex);
+	memset(tex, 0, sizeof(*tex));
+	// salt always > 0, so deleted texture will have illegal salt
+	tex->salt = 0; 
+	
+	return 1;
+}
 static void glnvg__dumpShaderError(GLuint shader, const char* name, const char* type)
 {
 	GLchar str[512+1];


src/nanovg_gl.h
```diff
@@ -345,12 +346,12 @@ static void glnvg__blendFuncSeparate(GLNVGcontext* gl, const GLNVGblend* blend)
...
+ 	GLNVGtexture* tex = NULL;
+	int i, idx = -1, salt;
 
 	for (i = 0; i < gl->ntextures; i++) {
 		if (gl->textures[i].id == 0) {
-			tex = &gl->textures[i];
+			idx = i;
 			break;
 		}
 	}
```
cforfang: ``tex == NULL`` below this line should probably be ``idx == -1`` now?

src/nanovg_gl.h
```diff
@@ -363,38 +364,54 @@ static GLNVGtexture* glnvg__allocTexture(GLNVGcontext* gl)
 			gl->textures = textures;
 			gl->ctextures = ctextures;
 		}
-		tex = &gl->textures[gl->ntextures++];
+		idx = gl->ntextures++;
```
cforfang: Instead of using ``rand()`` for the salt below, if you're here you know you're initializing a new entry for which you can set salt = 0.

src/nanovg_gl.h
```diff
@@ -147,6 +147,7 @@ struct GLNVGtexture {
 	int width, height;
 	int type;
 	int flags;
+	unsigned int salt;
```
cforfang: Isn't salt already stored as part of the id?

cforfang:
> Right-shifting & left-shifting was messing up on ints.

I didn't think this was a problem, but then I also don't use bitfields much :) What problems did you encounter?

> it appears that you were taking it from uninitialized memory, which isn't a reliable way to get unique salt for each texture

This seems like a bug in @memononen comment. I think you only need to identify when a new entry is allocated and set the salt to 0 then. If you add that I think the original code (which increments the salt on delete) should work.

Also, from skimming the code it's not clear to me why a separate ``salt`` field is needed. Is it not sufficent to just store it as part of ``id``?

src/nanovg_gl.h
Chaircrusher: Well... thinking about it.  The goal is to make the ID == the index in the array, so storing the ID in the Texture structure is no longer necessary.

The salt is what need to be stored.  The ID returned to the application is the ID + the salt.

src/nanovg_gl.h
memononen: That's correct, below should be:
```C
if (idx == -1) {
```

src/nanovg_gl.h
```diff
@@ -345,12 +346,12 @@ static void glnvg__blendFuncSeparate(GLNVGcontext* gl, const GLNVGblend* blend)
 
 static GLNVGtexture* glnvg__allocTexture(GLNVGcontext* gl)
 {
-	GLNVGtexture* tex = NULL;
-	int i;
+ 	GLNVGtexture* tex = NULL;
+	int i, idx = -1, salt;
```
memononen: int i, idx = -1, salt = 1;

src/nanovg_gl.h
```diff
@@ -345,12 +346,12 @@ static void glnvg__blendFuncSeparate(GLNVGcontext* gl, const GLNVGblend* blend)
...
 
 	for (i = 0; i < gl->ntextures; i++) {
 		if (gl->textures[i].id == 0) {
-			tex = &gl->textures[i];
+			idx = i;
```
memononen:
```C
idx = i;
salt = gl->textures[i].salt; // Carry salt from previous item
```

src/nanovg_gl.h
```diff
@@ -363,38 +364,54 @@ static GLNVGtexture* glnvg__allocTexture(GLNVGcontext* gl)
 			gl->textures = textures;
 			gl->ctextures = ctextures;
 		}
-		tex = &gl->textures[gl->ntextures++];
+		idx = gl->ntextures++;
 	}
 
+	tex = &gl->textures[idx];
+	while(!(salt = (unsigned int)rand()))
```
memononen: Remove rand() stuff.

src/nanovg_gl.h
```diff
@@ -363,38 +364,54 @@ static GLNVGtexture* glnvg__allocTexture(GLNVGcontext* gl)
...
+		;
 	memset(tex, 0, sizeof(*tex));
-	tex->id = ++gl->textureId;
+	tex->salt = salt;
+	// salt value is in high 2 bytes, ID in low two bytes
+	tex->salt <<= 16;
```
memononen: Remove
tex->salt <<= 16;

src/nanovg_gl.h
```diff
@@ -363,38 +364,54 @@ static GLNVGtexture* glnvg__allocTexture(GLNVGcontext* gl)
...
+	tex->id = (idx & 0xffff) | tex->salt;
```
memononen: Maybe the bitshift problem was operator precedence, try:
```C
tex->id = (idx & 0xffff) | (tex->salt << 16);
```

src/nanovg_gl.h
```diff
@@ -363,38 +364,54 @@ static GLNVGtexture* glnvg__allocTexture(GLNVGcontext* gl)
...
+	if (gl->textures[idx].salt != salt)
+		return 0;
 
+	// Clear texture
+	tex = &gl->textures[idx];
+	if (tex->tex != 0 && (tex->flags & NVG_IMAGE_NODELETE) == 0)
+		glDeleteTextures(1, &tex->tex);
+	memset(tex, 0, sizeof(*tex));
+	// salt always > 0, so deleted texture will have illegal salt
+	tex->salt = 0;
```
memononen: Salt must be increment here, not set to zero.

src/nanovg_gl.h
```diff
@@ -363,38 +364,54 @@ static GLNVGtexture* glnvg__allocTexture(GLNVGcontext* gl)
...
-		if (gl->textures[i].id == id) {
-			if (gl->textures[i].tex != 0 && (gl->textures[i].flags & NVG_IMAGE_NODELETE) == 0)
-				glDeleteTextures(1, &gl->textures[i].tex);
-			memset(&gl->textures[i], 0, sizeof(gl->textures[i]));
-			return 1;
-		}
-	}
-	return 0;
-}
+	int idx = id & 0xffff, salt = ((unsigned int)id) & 0xffff0000;
```
memononen: The bitshift should work here too no problem.

Chaircrusher:
_I didn't think this was a problem, but then I also don't use bitfields much :) What problems did you
encounter?_

The problems was using ints. When you right-shift a signed integer, it propogates the sign bits left.  This means that for some numbers X << 16 != (X << 16) >> 16.

I solved the problem by just making the salt a random integer left-shifted into the high two bytes.

_This seems like a bug in @memononen comment. I think you only need to identify when a new entry is allocated and set the salt to 0 then. If you add that I think the original code (which increments the salt on delete) should work._

Two things: 1) the existing code thinks an ID of zero means allocTexture failed, which means ID can never be zero. 2) The salt could be an incrementing counter, but it should not start at zero, because the index can be zero, so 0 + 0 violates assumption #1 

_Also, from skimming the code it's not clear to me why a separate salt field is needed. Is it not sufficent to just store it as part of id?_

As I said above, the ID is no longer needed as part of the Texture struct.  

I'll submit a new version of this branch that A) takes out the id from the Texture struct and B) uses monotonically increasing salt values.

Chaircrusher:
working on a new version.  will update within the hour.

cforfang:
> The problems was using ints. When you right-shift a signed integer, it propogates the sign bits left. This means that for some numbers X << 16 != (X << 16) >> 16.

But is that a problem in this case? I know left shifts of signed integers is undefined, but I wasn't sure if right-shifts for the purpose of extracting two numbers like here is OK.

src/nanovg_gl.h
Chaircrusher: The salt as a local variable in allocTexture means that it isn't preserved across calls. I made it a member of GLNVGcontext as textureSalt.

Chaircrusher:
I checked in a new version. I think that it is simple and correct.  I've verified that it works properly in the application I was testing in (vcvRack).  If you want to do it differently be my guest.

src/nanovg_gl.h
Chaircrusher: done

src/nanovg_gl.h
```diff
@@ -363,36 +366,52 @@ static GLNVGtexture* glnvg__allocTexture(GLNVGcontext* gl)
...
-		tex = &gl->textures[gl->ntextures++];
+		idx = gl->ntextures++;
 	}
 
+	tex = &gl->textures[idx];
 	memset(tex, 0, sizeof(*tex));
-	tex->id = ++gl->textureId;
+	// pre-increment so salt begins at one. When GLNVGcontext is
+	// allocated initially it is set to 0 via memset.
+	tex->salt = ++gl->textureSalt;
```
memononen: The reason for using a salt per texture entry is that they are less likely to wrap around and keep unique. I'd prefer to use that version.

PR #459, opened 2018-02-23
Feature: Catmull-Rom curves

c-mendoza:
New command nvgCurveTo to create Catmull-Rom splines. C-R splines are sometimes easier to use that beziers because they don't require you to supply control points.
Added a "curveResolution" property to NVGState to control the amount of detail that the curves are rendered with.
Added a "nvg_tesselateCurve" function that creates the points for the curve.

### DIFF
--- a/src/nanovg.c
+++ b/src/nanovg.c
@@ -55,6 +55,7 @@ enum NVGcommands {
 	NVG_BEZIERTO = 2,
 	NVG_CLOSE = 3,
 	NVG_WINDING = 4,
+	NVG_CURVETO = 5
 };
 
 enum NVGpointFlags
@@ -83,6 +84,7 @@ struct NVGstate {
 	float fontBlur;
 	int textAlign;
 	int fontId;
+	int curveResolution;
 };
 typedef struct NVGstate NVGstate;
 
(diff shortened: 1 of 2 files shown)


src/nanovg.c
```diff
@@ -1352,6 +1394,17 @@ static void nvg__flattenPaths(NVGcontext* ctx)
...
 			i += 3;
 			break;
+			case NVG_CURVETO:
+				curveCommands++;
+				p = &ctx->commands[i+1];
+				if (curveCommands > 3) {
+					cp1 = &ctx->commands[i - (3*3) + 1];
+					cp2 = &ctx->commands[i - (2*3) + 1];
+					cp3 = &ctx->commands[i - (1*3) + 1];
+					nvg__tesselateCurve(ctx, cp1[0], cp1[1], cp2[0], cp2[1], cp3[0], cp3[1], p[0], p[1], NVG_PT_CORNER);
```
kritzikratzi: if i understand correctly this requires four curveto() commands to do anything. imho it would be good if you allowed the special case where curveCommands>2 and you use nvg__lastPoint() for the first point instead. this would allow move()curve()curve()curve(). 


in general, have you considered adding in the curve parameter to allow also distance parametrized and chordal variants of catmul rom? 

the cp1...3 names are misleading here imho. you're doing this to save variables?

PR #469, opened 2018-03-29
Use GLES2 glTex[Sub]Image2D code paths also for GL2

### LINKED ISSUE #468: GL_RED is not a valid internalFormat in OpenGL 2.1
These lines assume that GL2 supports GL_RED, which is actually only in core since GL3.0 (and GL_EXT_texture_rg in GLES land).
GL_LUMINANCE is available since GL2.1, but disappears in OpenGL 4 (which should be covered by the GL3 cases).

https://github.com/memononen/nanovg/blob/e70a78e42e781bc73393afe47a62e6c049197a0b/src/nanovg_gl.h#L757-L763

First case should be for GLES2 and GL2. #else case will only with GL3

https://github.com/memononen/nanovg/blob/e70a78e42e781bc73393afe47a62e6c049197a0b/src/nanovg_gl.h#L849-L853

Same here, second case doesn't consider GL2.

strfry:
Fixes #468


memononen:
Thanks for the fix!

PR #477, opened 2018-04-26
Fixes the type of view size.

olliwang:
This commit changes the type of the view size accepted by `nvgRenderFrame()` from `int` to `float` so it matches the type defined in `GLNVGcontext`. This would prevent a framebuffer missing a pixel in final rendering result.

### DIFF
--- a/src/nanovg.c
+++ b/src/nanovg.c
@@ -363,7 +363,7 @@ void nvgDeleteInternal(NVGcontext* ctx)
 	free(ctx);
 }
 
-void nvgBeginFrame(NVGcontext* ctx, int windowWidth, int windowHeight, float devicePixelRatio)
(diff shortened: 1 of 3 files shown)


olliwang:
@memononen Can you review this PR? In iOS, the `devicePixelRatio` is usually an integer, that's fine. However, in some Android phone, like my current development device Nexus 5X. The `devicePixelRatio` is `2.6525`, and the screen resolution is 1080 x 1794, which means the `windowWidth` and `windowHeight` is `411.4285714286` and `683.4285714286`, respectively. If `nvgBeginFrame()` takes only integer size, that would cause visual problem.

memononen:
Merged. Thans for the improvement!

I was about to ask why float dimension is needed, thanks for the clarification, makes sense totally :)

olliwang:
@memononen Thank you very much. :)

PR #481, opened 2018-05-17
fix error handling

lieff:
ctx can't be != NULL on error path, "gl" is leaked.


src/nanovg.c
```diff
@@ -288,7 +288,11 @@ NVGcontext* nvgCreateInternal(NVGparams* params)
 	FONSparams fontParams;
 	NVGcontext* ctx = (NVGcontext*)malloc(sizeof(NVGcontext));
 	int i;
-	if (ctx == NULL) goto error;
+	if (ctx == NULL) {
```
memononen: This should be (ctx->params is not set yet either):
```C
if (ctx == NULL) {
	// Call renderer delete explicitly. Context is not initialized yet, so we cannot call nvgDeleteInternal() which usually calls the function.
	if (params->renderDelete != NULL)
		params->renderDelete(params->userPtr);
	return  NULL;
}
```

src/nanovg_gl.h
```diff
@@ -1590,7 +1590,6 @@ NVGcontext* nvgCreateGLES3(int flags)
 
 error:
 	// 'gl' is freed by nvgDeleteInternal.
```
memononen: remove comment

lieff:
Updated)

lieff:
Is changes good? Anything needed from my side?

PR #483, opened 2018-05-18
Make a --with-freetype option instead of having to uncomment

lehoff:
In premake4.lua there was a comment about what to uncomment when needing FreeType support.
This has now been turned into a premake option that can be passed in as a command line option like this:
```
$ premake4 --with-freetype gmake
```

### DIFF
--- a/premake4.lua
+++ b/premake4.lua
@@ -6,13 +6,22 @@ solution "nanovg"
 	configurations { "Debug", "Release" }
 	platforms {"native", "x64", "x32"}
 
+    newoption 
+    {
+        trigger = "with-freetype",
+        description = "Compile with FreeType support"
+    }
+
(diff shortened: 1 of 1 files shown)


cdegroot:
By the way, the reason for me for having to use Freetype is that when I tried NanoVG with the STB Truetype library on a RaspberryPi using VideoCore, strange things happened. The errors that came back indicated a failure to allocate memory, but looking at malloc traces that is unlikely, so it seems that in this particular configuration, the STB code writes in a wrong place. Freetype works fine (and I guess should render somewhat nicer, even if it enlarges the footprint of NanoVG a bit?). Currently I just forked NanoVG with a patched premake4.lue but having a command line option would be nicer. 

So consider this my: yes, please, accept this PR ;-)

PR #485, opened 2018-05-30
Create Nanovg.a

lpieri:
I create a new lib with nanovg and I the file now can compiling with flag -Wall -Wextra -Werror


vinjn:
@lpieri Can you make a new MR that still uses `premake` syntax?

PR #493, opened 2018-07-10
set extra glyph scale factor from freetype font metrics

kyriosli:
FreeType fonts has an internal scale factor that makes glyphs  smaller than the actual font size. So we should multiply the scale factor before calculating the glyph sizes.


olliwang:
Just tried and still found inconsistent font size displayed for different font files. Any way to improve it further?

PR #495, opened 2018-07-16
Fixes text corruption on atlas rebuild.

### LINKED ISSUE #494: Text corruption on Atlas rebuild.
I've run into an issue on NVidia hardware where text will get corrupted when the atlas needs to be rebuilt.  Either the characters will be missing or garbled looking.  I've tracked down and was able to fix the issue by changing 
Lines 2463-2468 in nanovg.c
			if (!nvg__allocTextAtlas(ctx))
				break; // no memory :(
			if (nverts != 0) {
				nvg__renderText(ctx, verts, nverts);
				nverts = 0;
			}

To 
			if (nverts != 0) {
				nvg__renderText(ctx, verts, nverts);
				nverts = 0;
			}
			if (!nvg__allocTextAtlas(ctx))
				break; // no memory :(

This is merely forcing the render of pending text before reallocating the Atlas.  Waiting till after reallocating the atlas causes the wrong texture data to be drawn for the characters in the old atlas.

olliwang:
This commit fixes #494. I've seen characters missing on iOS and Android couple of times. I'd like to help submitting PR for this fix. Thanks @jlinse.

### DIFF
--- a/src/nanovg.c
+++ b/src/nanovg.c
@@ -1729,7 +1729,7 @@ static int nvg__expandStroke(NVGcontext* ctx, float w, float fringe, int lineCap
 
 	w += aa * 0.5f;
 
-	// Disable the gradient used for antialiasing when antialiasing is not used. 
+	// Disable the gradient used for antialiasing when antialiasing is not used.
 	if (aa == 0.0f) {
 		u0 = 0.5f;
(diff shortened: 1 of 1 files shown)


memononen:
Thanks fo the fix!

PR #496, opened 2018-07-23
fix image repeation in gles2

kyriosli:
GLESv2 does not support for texture parameter GL_REPEAT when image size is non-power-of-two, but we can use fract method to normalize the texture coord to get the image repeated.

This only take effect when the drawing area size is larger than the image size.


memononen:
Good idea! We should only do this on GLES2, since it will leave a seam at the texture border when texture is magnified.

kyriosli:
That's right, so what we should do, add another macro which means we are using gles2 ?

PR #498, opened 2018-07-27
Forces autohint when loading glyph using FreeType.

olliwang:
This commit adds the `FT_LOAD_FORCE_AUTOHINT` parameter to `FT_Load_Glyph()` to fix wonky non-alphabetic  characters. 

I found this issue when I tried to display Google's open source Noto CJK font on an Android device. (CJK font combines Chinese, Japanese, and Korean fonts into one file because they all share a lot of Chinese and variant characters.), which is also the default font when setting Android's system language to Chinese.

This picture shows the big difference when adding `FT_LOAD_FORCE_AUTOHINT`. The font displayed on the left looks pretty bad and nothing like the font displayed by Android UI. However, the font displayed on the right looks much nature and matches the font displayed by Android UI.

nanovg_autohint

### DIFF
--- a/src/fontstash.h
+++ b/src/fontstash.h
@@ -204,7 +204,7 @@ int fons__tt_buildGlyphBitmap(FONSttFontImpl *font, int glyph, float size, float
 
 	ftError = FT_Set_Pixel_Sizes(font->font, 0, (FT_UInt)(size * (float)font->font->units_per_EM / (float)(font->font->ascender - font->font->descender)));
 	if (ftError) return 0;
-	ftError = FT_Load_Glyph(font->font, glyph, FT_LOAD_RENDER);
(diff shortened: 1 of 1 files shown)


olliwang:
@memononen Should I use a different method?

memononen:
Sorry, I read it and got distracted and forgot to get back.

It looks a lot better. I'm willing to merge the PR, but I'd like to see how it affects other glyphs. Especially smaller sizes. Could you post a screenshot of the demo with and without the hinting turned on?

olliwang:
Thanks for following up. Smaller sizes works better, too. I've also tested some western alphabet fonts and this parameter won't affect the rendering results. Currently, only Google's CJK font looks different in my test.

autohint

memononen:
Ok, merged. Thanks for the improvement!

olliwang:
Thanks. 👍

PR #503, opened 2018-08-11
Update README.md with another project link

sbarisic:
https://github.com/sbarisic/nanovg_dotnet


memononen:
Thanks for the link!

Another option for the API style could have been html canvas style:
```C#
NanoVGContext ctx = NVG.CreateContext(...);

ctx.BeginPath(...);
ctx.Fill(...);
```

sbarisic:
Thanks for the tip, i created a new ``NanoVGContext`` struct and made all of the functions as extension methods. So both API styles are now available. 😄

PR #506, opened 2018-09-09
Fixes the bug that gl may fail to draw fringes.

olliwang:
This commit fixes the bug that `glnvg__convexFill()` tried to draw fringes depending on whether `NVG_ANTIALIAS` is set. However, #386 adds `nvgShapeAntiAlias()` for chaning anti-alias state explicitly, and `glnvg__convexFill()` would cause error If the anti-alias state is disabled.


memononen:
Thanks for the fix!

PR #507, opened 2018-09-27
Fixes memory leaks of freetype.

olliwang:
This commit fixes the issue that freetype library isn’t released in `fonsDeleteInternal()` and causes memory leaks.


memononen:
Good catch! Thanks for the fix.

PR #508, opened 2018-09-27
Fix compile issue for missing { in fontstash.h

kennysharma:
master fails to build due to a missing `{` introduced by commit 9b27dc38947a69d2b93c9a3b6b5a707f159e9b2e.


memononen:
Thanks for the quick fix!

PR #516, opened 2018-12-14
Add render state saved count for restore

xiangwencheng1994:


xiangwencheng1994:
If merged, we can define "void nvgRestore(NVGcontext* ctx, int count = -1)" for nvgRestore function to compatible with old versions

PR #555, opened 2020-01-29
Sanitize demo.c and perf.c not to introduce a dependency on OpenGL or GLFW headers needlessly.

marcel303:
Sanitize demo.c and perf.c so they can be used to make a demo app with any backend, without introducing a dependency on OpenGL or GLFW headers needlessly.

### DIFF
--- a/example/demo.c
+++ b/example/demo.c
@@ -5,7 +5,9 @@
 #ifdef NANOVG_GLEW
 #  include <GL/glew.h>
 #endif
-#include <GLFW/glfw3.h>
+#ifdef DEMO_OPENGL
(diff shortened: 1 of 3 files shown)


marcel303:
I wrote a NanoVG backend for framework, which is available over here,
https://github.com/marcel303/framework/tree/master/3rdparty/nanovg-framework

And was running into the issue of not being able to compile an example app using demo.c and perf.c from NanoVG's repository. I put some OpenGL-only functionality behind a DEMO_OPENGL compile-time option. This makes demo.c and perf.c (and thus the example app) work with any backend, and without any OpenGL or GLTF dependency.

Issue #382, opened 2017-06-16
Set the same size, use different ttf file, the text size of the display difference


tailangjun:
In order to show Simplified Chinese, I tried several different ttf files. I found that set the same size, use different ttf file, the text size of the display difference. Should we need to read out some parameters from ttf file and set to fontstash?
华文黑体.ttf is bigger than UILabel, use the same fontsize.
PingFang Regular.ttf is smaller than UILabel, use the same fontsize.

1.use PingFang Regular.ttf
![usepingfang](https://user-images.githubusercontent.com/12316965/27180156-8b0ff10c-5204-11e7-8c4f-3383991aaac0.jpeg)

2.华文黑体.ttf
![usehuawen](https://user-images.githubusercontent.com/12316965/27180181-b239e512-5204-11e7-93d8-8849bbbc9397.jpeg)

memononen:
Can you post the values of `fonsVertMetrics()` for both fonts?

tailangjun:
1.use PingFang Regular.ttf
<img width="329" alt="pingfang" src="https://user-images.githubusercontent.com/12316965/27259139-738bd672-543e-11e7-8c05-974e687d3349.png">
2.use 华文黑体.ttf
<img width="315" alt="default" src="https://user-images.githubusercontent.com/12316965/27259141-7abafebe-543e-11e7-97e2-42888a8abcb3.png">

memononen:
Looks like there's two ways to calculate the ascender and descender, one used by OSX and another one by Windows (of course!): https://stackoverflow.com/questions/19856094/computing-font-truetype-hhea-values-ascender-descender

`stb_truetype` (font rasterizer used by nanovg) is using the OSX variant. Can you try to change the following functions in stb_truetype.h, to use the windows variant. 

```C
STBTT_DEF void stbtt_GetFontVMetrics(const stbtt_fontinfo *info, int *ascent, int *descent, int *lineGap)
{
    if (info->os2 != 0 && ttSHORT(info->data+info->os2 + 0) > 0) { // version > 0
        if (ascent ) *ascent  = ttSHORT(info->data+info->os2 + 68); // sTypoAscender
        if (descent) *descent = ttSHORT(info->data+info->os2 + 70); // sTypoDescender
        if (lineGap) *lineGap = ttSHORT(info->data+info->os2 + 72); // sTypoLineGap
    } else {
        if (ascent ) *ascent  = ttSHORT(info->data+info->hhea + 4);
[...]

PR #558, opened 2020-02-10
Fixes the issue that font size is inconsistent.

olliwang:
This commit fixes the issue that font size is inconsistent between different font files. `stb_truetype.h` is also updated to the latest version to support additional font types such as OpenType fonts.

The inconsistent font size issue was mentioned at #382 and I experienced the same issue for my app a long time ago. I always see different font sizes (even if they set to the same value) with different font files on screen.

Here is the current NanoVG's implementation result compared to Apple's Pages app. There are three font files used for English / Chinese / Thai, respectively.

 * English: Gotham Round (TTF)
 * Chinese: Noto Sans CJK (OTF)
 * Thai: Noto San Thai (TTF)

NanoVG Font Test 1

The Pages app shows bolder font for some reason, but that doesn't matter. Here we can see the stb backend can't render OpenType fonts. That's because NanoVG uses a very old `stb_truetype.h` by now. After updating the header file to the latest version. Both `stb` and `FreeType` backends show the same result.

NanoVG Font Test 2

However, the font sizes are still inconsistent. Only English character's font size is correct. After some investigation of `fontstash.h`. I found that this was caused by `fons__tt_getPixelWidthScale()`. The ascender and descender are used to calculate the font height and thus the scale. However, for some reason, the calculated font height diffs a lot in different font files. After some experiments, I managed to fix it by calculating the scale using `unit per em`. Now the font size is consistent across different font files.

NanoVG Font Test 3

### DIFF
--- a/src/fontstash.h
+++ b/src/fontstash.h
@@ -194,7 +194,7 @@ void fons__tt_getFontVMetrics(FONSttFontImpl *font, int *ascent, int *descent, i
 
 float fons__tt_getPixelHeightScale(FONSttFontImpl *font, float size)
 {
-	return size / (font->font->ascender - font->font->descender);
+	return size / font->font->units_per_EM;
 }
 
 int fons__tt_getGlyphIndex(FONSttFontImpl *font, int codepoint)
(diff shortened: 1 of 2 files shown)


memononen:
Thanks @olliwang, excellent work as always!

olliwang:
You are very welcome. Cheers! :D

PR #559, opened 2020-02-17
Support for loading a certain font face from a font file.

olliwang:
This commit allows loading a certain font face from a font collection file such as TTC. 

TTC file is very important for CJK fonts. For example, Android includes `.ttc` font files for CJK characters due to the fact that Chinese/Japanese/Korean share a lot of characters. Without a `.ttc` file, we need four files (each file takes at least 15 MB) to support all of Japanese/Chinese/Korean characters. The Noto-CJK TTC font file included in Android only takes ~25 MB.

Currently, the stb implementation in NanoVG does not support TTC at all as mentioned in #554. And the FreeType implementation only loads the first font face in the TTC. This isn't ideal as the same character may look different in different languages.

cjk

Without TTC support, we need to bundle 4 font files to cover all CJK characters, and we can't use the `.ttc` file included in Android system already. That would make the app super heavy.

### DIFF
--- a/src/fontstash.h
+++ b/src/fontstash.h
@@ -102,8 +102,8 @@ int fonsExpandAtlas(FONScontext* s, int width, int height);
 int fonsResetAtlas(FONScontext* stash, int width, int height);
 
 // Add fonts
-int fonsAddFont(FONScontext* s, const char* name, const char* path);
-int fonsAddFontMem(FONScontext* s, const char* name, unsigned char* data, int ndata, int freeData);
+int fonsAddFont(FONScontext* s, const char* name, const char* path, int fontIndex);
+int fonsAddFontMem(FONScontext* s, const char* name, unsigned char* data, int ndata, int freeData, int fontIndex);
 int fonsGetFontByName(FONScontext* s, const char* name);
 
 // State handling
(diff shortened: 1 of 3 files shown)


src/fontstash.h
```diff
@@ -102,8 +102,8 @@ int fonsExpandAtlas(FONScontext* s, int width, int height);
 int fonsResetAtlas(FONScontext* stash, int width, int height);
 
 // Add fonts
-int fonsAddFont(FONScontext* s, const char* name, const char* path);
-int fonsAddFontMem(FONScontext* s, const char* name, unsigned char* data, int ndata, int freeData);
+int fonsAddFont(FONScontext* s, const char* name, const char* path, int offset);
```
memononen: Based on freetype and stb_tt, I think we should call this `fontIndex`, not offset.

src/nanovg.h
```diff
@@ -545,10 +545,12 @@ void nvgStroke(NVGcontext* ctx);
 // Creates font by loading it from the disk from specified file name.
```
memononen: offset -> fontIndex, add comment:
// fontIndex specifies which font to load from a .ttf/.ttc file.

src/fontstash.h
```diff
@@ -278,13 +278,14 @@ int fons__tt_done(FONScontext *context)
...
-int fons__tt_loadFont(FONScontext *context, FONSttFontImpl *font, unsigned char *data, int dataSize)
+int fons__tt_loadFont(FONScontext *context, FONSttFontImpl *font, unsigned char *data, int dataSize, int index)
 {
-	int stbError;
+	int offset, stbError;
 	FONS_NOTUSED(dataSize);
 
 	font->font.userdata = context;
-	stbError = stbtt_InitFont(&font->font, data, 0);
+	offset = stbtt_GetFontOffsetForIndex(data, index);
```
memononen: Should error if offset is -1

olliwang:
@memononen Thanks. I just updated the code. 👍

memononen:
Thanks or the improvement!

olliwang:
Thanks for merging. Cheers! :)

PR #560, opened 2020-02-25
Support for resetting fallback fonts.

olliwang:
This commit adds `nvgResetFallbackFontsId()` and `nvgResetFallbackFonts()` functions for resetting fallback fonts. This would be very useful for improved CJK font support.

For example. If we always use an English-only font as the base font. We need to make sure the first fallback font is a Japanese font for Japanese articles or a Chinese font for Chinese articles. Right now there is no way to reset fallback fonts and thus the first assigned CJK font will be used across different languages.

### DIFF
--- a/src/fontstash.h
+++ b/src/fontstash.h
@@ -797,6 +797,17 @@ int fonsAddFallbackFont(FONScontext* stash, int base, int fallback)
 	return 0;
 }
 
+void fonsResetFallbackFont(FONScontext* stash, int base)
+{
+	int i;
+
+	FONSfont* baseFont = stash->fonts[base];
+	baseFont->nfallbacks = 0;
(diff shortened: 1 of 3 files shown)


src/fontstash.h
```diff
@@ -797,6 +797,17 @@ int fonsAddFallbackFont(FONScontext* stash, int base, int fallback)
 	return 0;
 }
 
+void fonsResetdFallbackFont(FONScontext* stash, int base)
```
memononen: Remove extra D:
fonsReset*d*FallbackFont

olliwang: Oops! Just fixed the typo. :)

memononen:
Thanks again for the improvement!

PR #561, opened 2020-02-26
Pass fringe parameter to renderTriangles

olilarkin:
fixes bad clipping of text on high DPI screens

Here is an example of the results we are getting without this fix, the text is clipped to control bounds with nvgscissor, but if the text goes outside the bounds we get some artifacts at the edges on retina screens. On low-DPI screens there are no such artifacts.

nvgfringe


olilarkin:
@olliwang we also needed to update this in MetalNanoVG, will do a PR there too

memononen:
Thanks for the fix!

PR #563, opened 2020-02-28
Fixes the issue that line gap may ruin the text layout.

olliwang:
iOS uses the `Hiragino Sans` font as the default Japanese font. However, this font abuses the `line gap` property like many fonts.

 * Adobe fonts usually treat `line gap` as the `line height`.
 * Most other western fonts sets `line gap` to 200.

Back to the `Hiragino` font. The font family always sets `line gap` to 500 no matter what while the `font height (ascent - descent)` is only 1000. Which causes the spaces between lines are super big. 

Screen Shot 2020-02-28 at 5 31 10 PM

As mentioned here https://stackoverflow.com/questions/5414730/custom-installed-font-not-displayed-correctly-in-uilabel, Apple ignores the `line gap` property since iOS 7, and a workaround to fix this issue is to add the `line gap` to the `ascent` and we'll be fine to ignore the `line gap` property. Now the result looks much normal.

Screen Shot 2020-02-28 at 5 30 27 PM

This change doesn't affect the alignment, either.

linegap

### DIFF
--- a/src/fontstash.h
+++ b/src/fontstash.h
@@ -953,10 +953,11 @@ int fonsAddFontMem(FONScontext* stash, const char* name, unsigned char* data, in
 	// Store normalized line height. The real line height is got
 	// by multiplying the lineh by font size.
 	fons__tt_getFontVMetrics( &font->font, &ascent, &descent, &lineGap);
+	ascent += lineGap;
 	fh = ascent - descent;
 	font->ascender = (float)ascent / (float)fh;
 	font->descender = (float)descent / (float)fh;
-	font->lineh = (float)(fh + lineGap) / (float)fh;
+	font->lineh = font->ascender - font->descender;
 
 	return idx;
 


memononen:
How will this affect how western multiline text is rendered? i.e. the nanovg demo.

olliwang:
This doesn't affect western multiline at all.

Screen Shot 2020-02-28 at 6 09 22 PM

PS. In my previous PR of fixing font size, the font size now matches what you saw on other native apps. The font used by the demo was affected by this because the font sizes were set to a value bigger than they actually were. I'm still fixing the font sizes and will submit another PR later.

olliwang:
Just submitted a PR to fix demo's font size. #564

The result looks exactly the same with or without this PR's changes of fixing line gap issue.

BEFORE:
Screen Shot 2020-02-28 at 6 22 05 PM

AFTER:
Screen Shot 2020-02-28 at 6 31 23 PM

memononen:
Thanks for the fix!

olliwang:
Thanks for the quick merging. Cheers! 🥂

PR #564, opened 2020-02-28
Fixes demo’s font sizes.

olliwang:
This commit fixes the demo’s font sizes for the changes in #558, which makes the rendered font size matches what you saw on other native apps. The demo used bigger font sizes than they actually were. This commit deduces the font sizes to match the original demo’s look.

Screen Shot 2020-02-28 at 6 22 05 PM


memononen:
Thanks for the quick fix.

Issue #398, opened 2017-08-05
Add infrastructure to compile with emscripten


rconde01:
If you did this, nanovg could be used as the basis of a web based dashboard...or lots of web based stuff I suppose.

Kerndog73:
The NanoVG API is based on the HTML5 Canvas API. 

https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D

Using the canvas directly would be faster than rasterizing in JavaScript and drawing that onto a canvas.

anlumo:
I've compiled nanovg with emscripten. It's pretty trivial, here is my makefile:

```Makefile
[...]

PR #567, opened 2020-03-02
Bind a dummy texture in setUniforms when compiling with emscripten

olilarkin:
fixes the WebGL error "RENDER WARNING: there is no texture bound to the unit 0"" in Chrome

note: previous PR worked with nanovg example, but not with iPlug2, where webgl context is initially hidden. In this PR i create the dummy texture on first use which seems to work in both cases

see discussion https://github.com/memononen/nanovg/issues/398#issuecomment-437004233

### DIFF
--- a/src/nanovg_gl.h
+++ b/src/nanovg_gl.h
@@ -704,13 +704,13 @@ static int glnvg__renderCreate(void* uptr)
 #endif
 	gl->fragSize = sizeof(GLNVGfragUniforms) + align - sizeof(GLNVGfragUniforms) % align;
 
+#ifdef __EMSCRIPTEN__
+	gl->dummyTex = glnvg__renderCreateTexture(&gl, NVG_TEXTURE_ALPHA, 1, 1, 0, NULL);
+#endif
+
 	glnvg__checkError(gl, "create done");
 
 	glFinish();
-
-#ifdef __EMSCRIPTEN__
-	gl->dummyTex = 0;
-#endif
 
 	return 1;
 }
@@ -994,9 +994,6 @@ static void glnvg__setUniforms(GLNVGcontext* gl, int uniformOffset, int image)
 		glnvg__checkError(gl, "tex paint tex");
 	} else {
 #ifdef __EMSCRIPTEN__
-		if(gl->dummyTex == 0) {
-			gl->dummyTex = glnvg__renderCreateTexture(&gl, NVG_TEXTURE_ALPHA, 2, 2, 0, NULL);
-		}
 		GLNVGtexture* tex = glnvg__findTexture(gl, gl->dummyTex);
 		glnvg__bindTexture(gl, tex->tex);
 #else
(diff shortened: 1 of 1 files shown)


olilarkin:
nope, this is no good either, closing

memononen:
Please update the PR instead of creating new one and closing each time. The method from MetalNanoVG should work just fine, i.e. create/find the texture in `renderCreate` and pass it in in `setUniforms`. What does not work for you?

olilarkin:
sorry about that. first PR worked (no more webgl warning) with emscriptenized nanovg demo, but not with iPlug2, where the webgl context is not yet onscreen at time . second PR worked with a basic iPlug2 project, but in this more complex demo https://iplug2.github.io/NANOVG/IPlugControls/ there were some new errors "[.WebGL-0x7fa3620ef200]GL ERROR :GL_INVALID_OPERATION : glDrawArrays: Source and destination textures of the draw are the same." And it looked totally wrong, which is why i closed the PR:
Screenshot 2020-03-05 at 20 15 21

memononen:
`glDrawArrays: Source and destination textures of the draw are the same` sounds like you ar rendering to a texture which you are using as a texture. Should not be related to this at all.

You have to have valid opengl context available when `renderCreate` is called, so creating the texture there should be valid too.

olilarkin:
Not sure what I had done to mess it up, but this is now working with both the nanovg demo and my project, so i have modified and reopened this PR

memononen:
Cool! I suggest following changes, I think it is safe to enable this on all cases.

Comment:
```C
...
	// Some platforms does not allow to have samples to unset textures.
	// Create empty one which is bound when there's no texture specified.
	gl->dummyTex = glnvg__renderCreateTexture(&gl, NVG_TEXTURE_ALPHA, 1, 1, 0, NULL);
```

Change logic to handles missing image (should not happen!):
```C
static void glnvg__setUniforms(GLNVGcontext* gl, int uniformOffset, int image)
{
	GLNVGtexture* tex = NULL;

	...

	if (image != NULL) {
		tex = glnvg__findTexture(gl, image);
	}
	// If not image is set, use empty texture.
	if (tex == NULL) {
		tex = glnvg__findTexture(gl, gl->dummyTex);
	}
	glnvg__bindTexture(gl, tex != NULL ? tex->tex : 0);
	glnvg__checkError(gl, "tex paint tex");
```

olilarkin:
I tried your changes, but the call to glnvg__renderCreateTexture() crashes with EXC_BAD_ACCESS in glnvg__allocTexture() on macOS.

memononen:
@olilarkin The only change to the glnvg__renderCreateTexture() bit was the comment. On second look, in that call it should be `gl` not `&gl`.

olilarkin:
ahha, thats it. pushed some updates

memononen:
Thanks for the fix!

Issue #373, opened 2017-05-26
nvgTextBreakLines() Incorrect layout Chinese.


tailangjun:
In function nvgTextBreakLines(), row->width is longer than breakRowWidth when display Chinese.
e.g. I display a Chinese string, breakRowWidth = 100, but the row->width = 104. 
It's maybe a bug.

Issue #206, opened 2015-03-12
Possible nvgTextBox and nvgTextBoxBounds inconsistency


SmilyOrg:
I'm not entirely certain where the blame lies, but I'm implementing text rendering for a framework and it's measuring the text first before rendering it with the measured width.

The problem is that either `nvgTextBoxBounds` seems to return a narrower width than `nvgTextBox` uses internally or that `nvgTextBox` is using wider or offset widths. With the Droid Sans font at size 27 and text "Button" I'm seeing `nvgTextBoxBounds` return 70 as the width, but setting that as the width while rendering with `nvgTextBox` it seems to come up with a final width of 71, which makes it break at "n", so it looks like this:

![2015-03-12_19-00-44](https://cloud.githubusercontent.com/assets/1451391/6624531/2d2cee3c-c8ea-11e4-9c83-feed55223c5b.png)

I've tried debugging it, but I haven't made much progress, however I did find that there are two different measurements `width` and `minx` / `maxx`. The former seems to match the 71 measurement and I think it's derived from `FONStextIter` measurements, while the latter seem to be derived from `FONSquad` ones, both provided by `fonsTextIterNext`.

I can fix this by rendering with a width 1 wider than the one measured:

![2015-03-12_19-15-58](https://cloud.githubusercontent.com/assets/1451391/6624841/39a29822-c8ec-11e4-9343-0a3d953a5069.png)

I'm not sure if that'll work for all cases and it doesn't seem like the right fix.

Any thoughts?

memononen:
They both use the same code to break the text into lines, but different break width can potentially give different results. I think row width is the correct value to use, not min/max. Min/max gives exact row bounds, while width is based on the text advance, the same is used for line breaking.

What if you add just a little bit, like 0.001f? Are you using centered text?

SmilyOrg:
Hmm, interesting. I'm using the default alignment (which is probably left aligned) and only adding a small epsilon doesn't work, since the conditional that ends up deciding the break
```c++
if (type == NVG_CHAR && nextWidth > breakRowWidth) {
```
ends up with `nextWidth` being 71 and `breakRowWidth` being 70 (the width that was measured) and presumably that makes it break at the last character.

memononen:
If you have debugger handy, can you check where the rowWidth gets set when you calculate the box size?

SmilyOrg:
First it's set in the `if (rowStart == NULL) { if (type == NVG_CHAR) { ...` conditional to 14.

[...]

PR #568, opened 2020-03-06
Fixes incorrect glyph width.

olliwang:
This commit fixes the bug that `nvgTextBounds()` may return incorrect results (#206) and `nvgTextBreakLines()` may give unmatched row widths (#373).

The bug is caused by determining quad’s `x1` by calculating the difference between a glyph’s `x0` and `x1`. However, glyph’s values lose some precision after converting to the integers for locating the glyph’s position in the atlas.

That’s the reason why there might be minor difference between the actual string width (NVGtextRow’s width) and the not always correct bound width (`bounds[2] - bounds[0]` returned by `nvgTextBounds()`). The former uses glyph’s `xadv` as each glyph’s width while the latter uses glyph’s `x1 - x0` to determine the width.

### DIFF
--- a/src/fontstash.h
+++ b/src/fontstash.h
@@ -70,6 +70,7 @@ struct FONSquad
 {
 	float x0,y0,s0,t0;
 	float x1,y1,s1,t1;
+	float xadv;
 };
 typedef struct FONSquad FONSquad;
 
@@ -1232,6 +1233,7 @@ static void fons__getQuad(FONScontext* stash, FONSfont* font,
 		q->y0 = ry;
 		q->x1 = rx + x1 - x0;
 		q->y1 = ry + y1 - y0;
+		q->xadv = glyph->xadv / 10.0f;
 
 		q->s0 = x0 * stash->itw;
 		q->t0 = y0 * stash->ith;
@@ -1245,6 +1247,7 @@ static void fons__getQuad(FONScontext* stash, FONSfont* font,
 		q->y0 = ry;
 		q->x1 = rx + x1 - x0;
 		q->y1 = ry - y1 + y0;
+		q->xadv = glyph->xadv / 10.0f;
 
 		q->s0 = x0 * stash->itw;
 		q->t0 = y0 * stash->ith;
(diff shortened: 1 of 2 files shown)


memononen:
Thanks for picking this up.

There are two widths for text, as I explained in #206. One for layouting the glyphs (width), and another one for the geometric bounds of the text (xmin/max).

What is the use case you've been testing?

If it problem is about how breaklines work, an alternative solution could be to change
https://github.com/memononen/nanovg/blob/master/src/nanovg.c#L2719
to:
```C
float nextWidth = q.x1 - rowStartX;
```

olliwang:
@memononen Thanks for the hint. I was implementing a text widget that can adjust the size automatically. I wasn't sure what's the correct way to get the width of a single line text.

I thought I should use `nvgTextBounds()`. In my case, the text `abcdefg` would give me `bounds[0] = 0` and `bounds[2] = 68`, so I tought the text width should be `68`.

However, if I passed `68` as the max width to `nvgTextBreakLines()`. It only draws `abcdef`, the `g` is missing. If I set the max width to a larger number such as 100 and check `NVGtextRow`'s `row` property, the value is `69`. The lack of space is the reason why `g` is dropped. I'm a bit confused of two different widths for my text case. I was trying to determine the width of a text, not glyphs. Is there other better way to get correct text width in addition to `nvgTextBounds()`?

memononen:
There was actually a bug in the minx/maxx calculations. I just fixed that.

Can you explain more how the dynamic adjustment works? I think I'm still missing something important.

If you draw text using nvgTextBox, then you should use nvgTextBoxBounds. Then there's no need to feedback the info.

The float value nvgTextBounds returns is the "logical width" of the text. nvgTextBoxBounds does not return such thing, maybe it could return max of all rows?

memononen:
@olliwang re: the coordinates, the bounds coordinates and row break width are kind of different units (bounding box of glyphs vs. advancement), so that's the slight discrepancy.

olliwang:
@memononen I didn't use `nvgTextBoxBounds()` because I don't know `breakRowWidth`. I'd like to explain more specific of what I was trying to do. 

I was implementing a button widget with a text widget inside. The button should always show one line string and fixed paddings around the text. So if the text is localized to different languages. The button (and text) width change accordingly with the same fixed paddings. 

To do this, I need to know the specific width of the text. Sure I can pass a super big number as `breakRowWidth` to `nvgTextBreakLines()` to get the correct `NVGtextRow`'s `width`, but I am not sure if it's the right way.

memononen:
If you have just one line of text, then you should use `nvgTextBounds()` to get the bounds and `nvgText()` to draw the text. That is, one should not generally mix text and text box stuff.

I still feel like there might be something that is obvious to me that I don't see in this case :)

olliwang:
@memononen You are right. `nvgTextBounds()` is enough for me. Thanks for your help. :)

PR #574, opened 2020-03-22
Vertical centering of text based on the ascender only

ghost:
I needed an option to center text vertically based on the cap height only. When choosing **NVG_ALIGN_MIDDLE**, the ascender and the descender are taken into account, so a string containing upper-case characters only, e.g., **TEXT**, is not centered vertically, but shifted upwards.

Therefore I added a new option **NVG_ALIGN_MIDDLE_ASCENT**, which is based in the height of the ascender only.

I think that cap height would be an even better option, but the fontstash library does not provide access to this measure.

### DIFF
--- a/src/fontstash.h
+++ b/src/fontstash.h
@@ -28,14 +28,15 @@ enum FONSflags {
 
 enum FONSalign {
 	// Horizontal align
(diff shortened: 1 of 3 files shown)


rgb2hsv:
I found a valuable use for this same feature. Thanks for PRing it!

PR #587, opened 2020-09-15
Added a preprocessor definition to disable the use of stb_image

christophe-f8:
I'm working on a large project with many libraries and applications and I use stb_image and/or nanovg in several of them.
Since stb_image is linked using a preprocessor definition (STB_IMAGE_IMPLEMENTATION) linked errors occurred when using both nanovg and some of my libraries using stb_image (attempt to link twice).

I made a small fix by disabling the use of stb_image in NanoVG by using a single preprocessor definition (NVG_NO_STB), which is undefined by default (i.e. STB is used by default), so this should have no effect at all for everyone, unless you manually add that definition.

Also, I've added a CMakeLists for building purposes.


memononen:
Hi! The changes to the nanovg.c looks good, but I cant accept the cmake
file as I cannot maintain it. Would you mind updating the PR to remove the
cmakelist.txt?

christophe-f8:
Sure, I removed CMakeLists.txt. It's pretty easy to write anyway should one need to.

I've updated my fork and I assume that it updates the PR as well, but let me know if I need to do something else (I'm really new to committing/pushing on GitHub)

memononen:
Merged, thanks for improvement!

PR #588, opened 2020-10-01
fix for freetype glyph loading 

d4tocchini:
With nanovg as renderer for @rxi's excellent [lite](https://github.com/rxi/lite) code editor, I noticed the following undesirable glyph wonkiness when using FreeType:

Screen Shot 2020-10-01 at 12 07 36 AM

The `s` glyph is particularly egregious.

Adding `FT_LOAD_TARGET_LIGHT` to the `FT_Load_Glyph` flags ([see docs](https://www.freetype.org/freetype2/docs/reference/ft2-base_interface.html#ft_load_target_xxx)) did the trick:

Screen Shot 2020-10-01 at 12 08 16 AM


memononen:
Thanks for the fix!

PR #603, opened 2020-12-25
fix #602

mulle-nat:
Sorry, previous PR needs a fix for #602.

PR #604, opened 2020-12-25
Continous Integration for nanvog

mulle-nat:
This PR adds a github workflow that builds nanovg (on ubuntu) for every push. Can be seen in action [here](https://github.com/mulle-nat/nanovg/actions).

This is useful, as PRs from inexperienced developers (ahem) that carry compilation errors will be caught early. This PR supersedes #603, as it also contains the compile fix. If it didn't then the CI would fail and superficially it would look, like it didn't work either. Which is something I wanted to avoid.

The CI does not build examples (yet), due to technical problems with the older ubuntu version being used by github.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,28 @@
+name: CI
+
+on: [push]
+
+env:
+  BUILD_TYPE: release
+
+jobs:
+  build:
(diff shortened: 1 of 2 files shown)


memononen:
Thanks for the fix and the CI script! I have not had the time to look into CI, but it does not look too intimidating :)

PR #615, opened 2021-05-26
Make data argument for nvgCreate*Mem() be pointer to const.

AlbertoGP:
This avoids compiler warnings when the source data is const.

The first change in line 32, removing the trailing space, is a stowaway but I think it’s not worth putting in a different commit.
I’ll remove that if asked to. Any other changes, please let me know.


AlbertoGP:
Ah, no, this just moves the warning upstream. It must not be const in the case when the library frees the data.

PR #620, opened 2021-10-31
Update fontstash.h

mulle-nat:
* change call order in `fonsDeleteInternal` so that free memory isn't passed to `fons__tt_done`


memononen:
Thanks for the fix!

Issue #610, opened 2021-03-03
Multi-context and multithread problems with FreeType


mulle-nat:
Nanovg uses a static `FT_Library` variable, which it initializes with `FT_Init_FreeType` during [`fons__tt_init`](https://github.com/memononen/nanovg/blob/master/src/fontstash.h#L162) which runs for each nanovg context being created. As soon as you are creating two nanovg contexts (and then destroying them), you will run into problems, as `FT_Init_FreeType` and `FT_Done_FreeType` don't [reference count](https://www.freetype.org/freetype2/docs/reference/ft2-base_interface.html#ft_library):
  
```c
static FT_Library ftLibrary;

int fons__tt_init(FONScontext *context)
{
	FT_Error ftError;
	FONS_NOTUSED(context);
	ftError = FT_Init_FreeType(&ftLibrary);
	return ftError == 0;
}

int fons__tt_done(FONScontext *context)
{
	FT_Error ftError;
	FONS_NOTUSED(context);
	ftError = FT_Done_FreeType(ftLibrary);
	return ftError == 0;
}
```

The [FreeType dox suggest](https://www.freetype.org/freetype2/docs/reference/ft2-base_interface.html#ft_init_freetype) replacing with `FT_New_Library` and `FT_Done_Library`.  These don't use [atomics for reference counting](https://github.com/freetype/freetype/blob/f9f6adb625c48ef15b5d61a3ac1709a068ea95a3/src/base/ftobjs.c#L5257) though and are therefore inherently not thread safe. So it's not a perfect solution, but solves the two context problem for a single thread app.
 
nanovg should either reference count (and skip) superflous `FT_Init_FreeType` and `FT_Done_FreeType` attempts or should embed the `FT_Library` static in the `FONScontext`. That would create duplicate fonts though.

Locking during nanovg context creation/deletion and font loading would still be needed for multi-threaded apps. I don't think locking would be a good addition to nanovg itself though.  

I will have to fix this for my project, but would like to get some feedback.

mulle-nat:
I have [implemented](https://github.com/mulle-kybernetik-tv/nanovg/tree/fix-truetype) `FT_Library` as a member of `FONScontext`. Not sure if this solves the multithread problems completely, but it should solve the multi-context problem. 
I will probably make a pull request, when this has been used a bit more.

AlexHarker:
There's a fix for this on navvy now, but I suspect it also entails us fixing the text sizes at last...

PR #621, opened 2021-12-02
Fixes FreeType errors when dealing with multiple NVGcontext objects.

olliwang:
Originally there is only a shared `FT_Library` object. However, if we create multiple `NVGcontext` objects and as soon as we delete one, the shared `FT_Library` object will be released, too. And that would cause a problem for the rest of `NVGcontext` objects.


memononen:
Thanks for the fix!

mulle-nat:
This should fix #610.

PR #622, opened 2022-01-05
Restore previous shader program on nvgEndFrame

clinthidinger:
This change restores the previous shader program instead of setting it to 0 on nvgEndFrame()/renderFlush().

This will also help when using nanovg with other graphics libraries that manage a default shader like Cinder.

### DIFF
--- a/src/nanovg_gl.h
+++ b/src/nanovg_gl.h
@@ -1180,7 +1180,7 @@ static void glnvg__renderFlush(void* uptr)
 {
 	GLNVGcontext* gl = (GLNVGcontext*)uptr;
 	int i;
-	
+
 	if (gl->ncalls > 0) {
 		GLuint lastProgram = 0;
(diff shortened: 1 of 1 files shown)


mulle-nat:
On one hand this seems like an improvement, but on the other hand isn't there a lot of other state clobbered by nvg ? Only partial restoration of the previous state may lead to other subtle or not so subtle bugs. Just my thoughts on this...

clinthidinger:
You're right about other states being clobbered.   I should have looked through the README more closely in the part where it describes the states that get touched.

However, I think it may be worth considering preserving the program state since it seems common for other graphics libraries to try to simulate a default shader.   The change to this state was actually causing an OpenGL error (GL_INVALID_OPERATION) from drawing with a null program handle when I called draw on a 3D object in Cinder after a call to nvgEndFrame().  It was also difficult to track down what the problem was until I used glEnable( GL_DEBUG_OUTPUT ).

PR #625, opened 2022-02-12
Fix stencil strokes and update to latest nanovg

nidefawl:
In this PR:
1. Merge latest nanovg

2. Fix Stencil Strokes: 
I added a new enum to handle the 3 different pipelines used for stencil strokes.  
I wanted to also use enums for the fill-path (nvgFill) to make the code cleaner, but haven't done that yet.

3. I combined the 2 CMakeLists.txt into example/CMakeLists.txt and added some bits to make it easier to build and debug in MSVC/VSCode.

4. I added GLAD as default gl-loader for the GL3 example. It can be switched back to GLEW (I kept glew to allow easier merge with main nanovg)

5. I Increased the number of swapchain images to minimum 3 and updated the swapchain barrier to gain some extra performance.

6. Spacebar renders the demo multiple times to add some 'load' when comparing performance

If you want to merge just bits of my branch you need to merge it yourself 😋


nidefawl:
Sorry, this was the wrong repo for the PR 🤣

PR #626, opened 2022-03-25
Fixes memory leaks caused by full font images.

olliwang:
This commit fixes the memory leak issue caused by not handling the situation where the number of allocated font images meets `NVG_MAX_FONTIMAGES`.

In this implementation of handling such situation. All existing font images will be released and a new one with the `NVG_MAX_FONTIMAGE_SIZE` will be created.

### DIFF
--- a/src/nanovg.c
+++ b/src/nanovg.c
@@ -28,7 +28,7 @@
 #ifndef NVG_NO_STB
 #define STB_IMAGE_IMPLEMENTATION
 #include "stb_image.h"
-#endif 
+#endif
 
 #ifdef _MSC_VER
 #pragma warning(disable: 4100)  // unreferenced formal parameter
@@ -396,6 +396,7 @@ void nvgEndFrame(NVGcontext* ctx)
 	ctx->params.renderFlush(ctx->params.userPtr);
 	if (ctx->fontImageIdx != 0) {
 		int fontImage = ctx->fontImages[ctx->fontImageIdx];
+		ctx->fontImages[ctx->fontImageIdx] = 0;
 		int i, j, iw, ih;
 		// delete images that smaller than current one
 		if (fontImage == 0)
@@ -404,20 +405,19 @@ void nvgEndFrame(NVGcontext* ctx)
 		for (i = j = 0; i < ctx->fontImageIdx; i++) {
 			if (ctx->fontImages[i] != 0) {
 				int nw, nh;
-				nvgImageSize(ctx, ctx->fontImages[i], &nw, &nh);
+				int image = ctx->fontImages[i];
+				ctx->fontImages[i] = 0;
+				nvgImageSize(ctx, image, &nw, &nh);
 				if (nw < iw || nh < ih)
-					nvgDeleteImage(ctx, ctx->fontImages[i]);
+					nvgDeleteImage(ctx, image);
 				else
-					ctx->fontImages[j++] = ctx->fontImages[i];
+					ctx->fontImages[j++] = image;
 			}
 		}
 		// make current font image to first
-		ctx->fontImages[j++] = ctx->fontImages[0];
+		ctx->fontImages[j] = ctx->fontImages[0];
 		ctx->fontImages[0] = fontImage;
 		ctx->fontImageIdx = 0;
-		// clear all images after j
-		for (i = j; i < NVG_MAX_FONTIMAGES; i++)
-			ctx->fontImages[i] = 0;
 	}
 }
 


src/nanovg.c
```diff
@@ -2416,8 +2416,21 @@ static int nvg__allocTextAtlas(NVGcontext* ctx)
 {
 	int iw, ih;
 	nvg__flushTextTexture(ctx);
-	if (ctx->fontImageIdx >= NVG_MAX_FONTIMAGES-1)
-		return 0;
+	if (ctx->fontImageIdx >= NVG_MAX_FONTIMAGES-1) {
+		int i;
```
memononen, line 2420: I wonder if we should call render flush here before deleting the images? I think now it's possible that because of the caching, there might be some render calls in flight when we delete the textures. That's the reason why the deletes happen at `nvgEndFrame()`.

olliwang:
@memononen Thanks for your hint. After examining the code again. I found that the memory leaks issue is actually caused by not releasing fog images properly in `nvgEndFrame()`.

For example, when the `NVG_MAX_FONTIMAGES` number of font images are allocated all at `NVG_MAX_FONTIMAGE_SIZE`. If the next frame use only 2 font images. `nvgEndFrame()` will set the last two font images to `0` directly without calling `nvgDeleteImage()`.

I've updated to code to initialize all font images to `0` in `nvgCreateInternal()` as well. (Currently this is done only in `nvgDeleteInternal()`. Also, whenever a font image is deleted in `nvgDeleteImage()`, the image handle resets to 0 immediately. That way, we can make sure the `clear all images after j` loop not reset image handle without releasing the corresponded image.

src/nanovg.c
```diff
@@ -330,6 +330,8 @@ NVGcontext* nvgCreateInternal(NVGparams* params)
 	ctx->fontImages[0] = ctx->params.renderCreateTexture(ctx->params.userPtr, NVG_TEXTURE_ALPHA, fontParams.width, fontParams.height, 0, NULL);
 	if (ctx->fontImages[0] == 0) goto error;
 	ctx->fontImageIdx = 0;
+	for (i = 1; i < NVG_MAX_FONTIMAGES; i++)
```
memononen, line 333: this is already done at line 298

src/nanovg.c
```diff
@@ -405,19 +408,25 @@ void nvgEndFrame(NVGcontext* ctx)
 			if (ctx->fontImages[i] != 0) {
 				int nw, nh;
 				nvgImageSize(ctx, ctx->fontImages[i], &nw, &nh);
-				if (nw < iw || nh < ih)
+				if (nw < iw || nh < ih) {
 					nvgDeleteImage(ctx, ctx->fontImages[i]);
-				else
+					ctx->fontImages[i] = 0;
+				} else {
 					ctx->fontImages[j++] = ctx->fontImages[i];
```
memononen, line 415: If this branch was:
```
int image = ctx->fontImages[i];
ctx->fontImages[i] = 0;
ctx->fontImages[j++] = image;
```
Then we should not need the loop at the bottom at all.

src/nanovg.c
```diff
@@ -405,19 +408,25 @@ void nvgEndFrame(NVGcontext* ctx)
...
 		// make current font image to first
 		ctx->fontImages[j++] = ctx->fontImages[0];
 		ctx->fontImages[0] = fontImage;
 		ctx->fontImageIdx = 0;
 		// clear all images after j
-		for (i = j; i < NVG_MAX_FONTIMAGES; i++)
-			ctx->fontImages[i] = 0;
+		for (i = j; i < NVG_MAX_FONTIMAGES; i++) {
+			if (ctx->fontImages[i] != 0) {
+				nvgDeleteImage(ctx, ctx->fontImages[i]);
```
memononen, line 426: I think this might currently remove too much. Imagine a case where we have 3 textures: A, B, C, and remove B. Then the array at this point will be A, C, C, and j=2 (after the remove it is A, 0, C, then C gets copied over). So we would delete C.

With the suggested change above, the array should be A, C, 0, after remove, and this loop is not needed at all.

memononen: Good catch on the leak!

src/nanovg.c
memononen: The copy is used so that `j == i` will not erase the array.

olliwang:
@memononen  Yeah, your solution is much better and cleaner. Just updated the code.  :)

src/nanovg.c
```diff
@@ -404,20 +405,21 @@ void nvgEndFrame(NVGcontext* ctx)
...
-				nvgImageSize(ctx, ctx->fontImages[i], &nw, &nh);
-				if (nw < iw || nh < ih)
-					nvgDeleteImage(ctx, ctx->fontImages[i]);
-				else
-					ctx->fontImages[j++] = ctx->fontImages[i];
+				int image = ctx->fontImages[i];
+				nvgImageSize(ctx, image, &nw, &nh);
+				if (nw < iw || nh < ih) {
+					nvgDeleteImage(ctx, image);
+					ctx->fontImages[i] = 0;
```
memononen, line 412: you can move this line right after `image = ...` as it's same on both branches.

memononen:
one more update, then we're good to go :)

olliwang:
Oops! How can I forget that. Just updated the code. :D

memononen:
Thank you for the fix!

olliwang:
Cheers!

nidefawl:
In my local version I implemented a handler for `fonsSetErrorCallback ` that reallocates the image on error code `FONS_ATLAS_FULL`.
This way I got rid of the `fontImages` array entirely.
And I was also able to remove these checks:
https://github.com/memononen/nanovg/blob/397f3300bc14bdb856ec429c71c345ab0340b0a6/src/nanovg.c#L2494-L2505

There was a bit more code that got a lot simpler that way

memononen:
The reason for the complexity is that some of the backends cache the data for a whole frame. So if you fiddle with the textures there, then you might render garbage.

PR #628, opened 2022-04-24
Direct opengl render

john-g-h-doe:
Something that was asked sometimes... if ever it is not yet here, very rough basis, sorry.

### DIFF
--- a/example/INSTALL.txt
+++ b/example/INSTALL.txt
@@ -1,0 +1,7 @@
+on linux, try this line to compile the demo :
+
+gcc -I../src/ ../src/nanovg.c example_gl2.c demo.c perf.c -lglfw -pthread -lGL -lGLU -lX11 -lXxf86vm -lXrandr -lpthread -lXi -ldl -lm -o demo
+
+you may need to :
+
+apt install libglfw3-dev libglu1-mesa-dev libxxf86vm-dev libxi-dev
--- a/example/demo.c
+++ b/example/demo.c
@@ -802,6 +802,34 @@ void drawLines(NVGcontext* vg, float x, float y, float w, float h, float t)
 	nvgRestore(vg);
 }
 
+static int directRender()
+{
+   // Draw a simple red triangle
+
+   // Color
+   glColor3f(1.0f, 0.0f, 0.0f);
+
+#define TRI_SIZE 0.1
+
+   // Vertices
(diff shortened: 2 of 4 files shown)


memononen:
Why would you need to do this?

john-g-h-doe:
I needed this several times before, and searched forums. Several other
persons wanted this, finally I gave up.
Now I feel I'll need it soon, so this time, this is here (just the basis).

This is for having for example easy bitmap display, text rendering, easy
and beautiful drawing, as well as all you can imagine to do with OpenGL
(ambient occlusion on 3d cubes, particle effects or anything).

mulle-nat:
This approach has the problem that nanovg caches the draw commands, so the OpenGL calls will appear before previously executed nanovg drawing commands in the same frame. I have no problem mixing OpenGL and nanovg, by using multiple `nvgBeginFrame` and `nvgEndFrame` calls in the same screen frame. I execute OpenGL after a `nvgEndFrame` and then resume my nanovg code with  `nvgBeginFrame`.

john-g-h-doe:
I was sure to have made the call of this "direct OpenGL rendering" code
after, needs more tests on my side. I will double-check soon, and will try
your method.

john-g-h-doe:
mulle-nat:
>This approach has the problem that nanovg caches the draw commands, so the OpenGL calls will appear before previously >executed nanovg drawing commands in the same frame. I have no problem mixing OpenGL and nanovg, by using multiple >nvgBeginFrame and nvgEndFrame calls in the same screen frame. I execute OpenGL after a nvgEndFrame and then resume >my nanovg code with nvgBeginFrame.

john-j-h-doe:
> I was sure to have made the call of this "direct OpenGL rendering" code after, needs more tests on my side. I will double-check soon, and will try your method.

@mulle-nat:
- double check done
- conclusion:
- you made a good joke too!
- <smile>

That was not voluntary on my side at first, this will make my day

mulle-nat:
Works for me though :man_shrugging: as in this demo, where the teapot is old OpenGL and all the rest is nanovg. 
opengl-mix 
The teapot draws over the lower bar and below upper bar. Expect no further help from me though.

memononen:
Multiple `nvgBeginFrame()`/`nvgEndFrame()` pairs or rendering to texture and using that within the nanovg is the way to go.

Issue #347, opened 2016-12-14
Separation of header and c files ?


ejahandar:
Hi, i can't figure out why nanovg_gl.h contains both definitions and declarations together, if i want to use nvgCreateGLES2 on multiple files so i have to include nanovg_gl.h (with -DNANOVG_GLES2_IMPLEMENTATION) in each files, now the problem occurs, there are so many files which have nvgCreateGLES2 compiled in and they're conflicting with each other in link time.
For the moment, i have separated definition and declarations in two separate file, am i correct?

kritzikratzi:
i think the idea is that you only have the implementation in one file, ie get rid of the `-DNANOVG_GLES2_IMPLEMENTATION`, then pick one cpp file where you do: 
 
    #define NANOVG_GLES2_IMPLEMENTATION
    #include "nanovg_gl.h"

every else you just include the header. (or... something like that :) )

ejahandar:
I agree with one file for implementation, but combining both header and source file will make problems if you need using nvgCreateGLES2 on multiple files say multiple surfaces.

memononen:
The one file idea is taken from stb libraries [1]. You should have only one file in your project where you define the implementation as @kritzikratzi mentioned.

[1] https://github.com/nothings/stb

holocronweaver:
Ran into the same problem. Perhaps this should be mentioned in the README? 

I am guessing the reason the stb approach is used is so GL lib wrangling is left up to user.

ejahandar:
what about using nvgluCreateFramebuffer (in nanovg_gl_utils.h) in multiple files, its impossible with this approach.

KrankerApfel:
Hello,

Sorry to reopen that old issue but I struggle a lot with the same problem.

Has anything changed regarding the inclusion of nanovg_gl.h in several .cpp ?
I tried to write in a first .cpp :

`
    #include <nanovg.h>
    #define NANOVG_GL3_IMPLEMENTATION
    #include <nanovg_gl.h>
`

There is only one .cpp file which define that as @kritzikratzi suggests and then in my other .cpp file, I only put : 

`     #include <nanovg.h>    
     #include <nanovg_gl.h>`

And this leads to the compile-time error   `C3861 : ncgCreateGL3  identifier not found`  from that second .cpp file.

i already tried to fix that by declaring the ncgCreateGL3 function  in my file without the definition, but the linker send a `LNK2019 : unresolved external symbol` for the definition.

Any idea of what is happening ?

[...]

PR #639, opened 2022-07-17
Add Get GL Texture Id

rgb2hsv:
This PR adds the ability to retrieve the GL texture Id for an image allocated by NVG. The idea is that we would want to map a frame buffer to an existing NVG image to render a shader with some type of animation. It's better to use an existing NVG texture because it preserves z-order with other NVG components.

### DIFF
--- a/example/demo.c
+++ b/example/demo.c
@@ -818,7 +818,7 @@ int loadDemoData(NVGcontext* vg, DemoData* data)
 			return -1;
 		} else {
 			const int glTextureId = nvgGetImageTextureId(vg, data->images[i]);
-			if(glTextureId<0) {
+			if(glTextureId < 0) {
 				printf("Invalid GL Texture Id for Image\n");
 			}
 		}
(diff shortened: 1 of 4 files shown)


mulle-nat:
There are already functions for this:

``` c
GLuint nvglImageHandleGLES2(NVGcontext* ctx, int image);
GLuint nvglImageHandleGLES3(NVGcontext* ctx, int image);
GLuint nvglImageHandleGL3(NVGcontext* ctx, int image);
GLuint nvglImageHandleGL2(NVGcontext* ctx, int image);
```

rgb2hsv:
That's cool! But those functions aren't public in nanovg.h header so that the application can use them easily. As far as I know, you can include nanovg.h in lots of files, but nanovg_gl.h only once. Their name is also specific to the implementation, which is not ideal.

mulle-nat:
Check #347 where I show, how to use `nanovg_gl.h` when not defining `NANOVG_..._IMPLEMENTATION`. I think the naming thing you can work around yourself fairly easily  (just my opinion though).

rgb2hsv:
Interesting .. sounds like a bit of a fluke related to the nvgCreateGL3 symbol being exported in one compilation and not in another because of how the defines are called (and compiler configured). The "include only once" behavior is an artifact of "header only" implementations as Mikko pointed out. Still think this PR will make things easier for anyone trying to add custom shaders in with their NVG (like me).

PR #640, opened 2022-07-25
Font Dilation (font outline)

rgb2hsv:
This PR adds font dilation as nvgFontDilate() to give better emphasis to text, especially when it is overlaid on an unknown background color or image. It is implemented with gray level morphology, which is approximated with directional convolutions. It can also be used in combination with nvgFontBlur() for a smoother fall-off.
dump

### DIFF
--- a/src/fontstash.h
+++ b/src/fontstash.h
@@ -1081,17 +1081,18 @@ static void fons__blur(FONScontext* stash, unsigned char* dst, int w, int h, int
 static void fons__maxRows(unsigned char* dst, int w, int h, int dstStride)
 {
 	int x, y;
+	unsigned char prev, current;
 	for (x = 0; x < w; x++) {
-		unsigned char prev=dst[0];
+		prev=dst[0];
 		for (y = dstStride; y < h*dstStride; y += dstStride) {
-			unsigned char current=dst[y];
+			current=dst[y];
 			if(prev > current){
 				dst[y]=prev;
 			}
 			prev=current;
 		}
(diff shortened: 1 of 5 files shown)


src/fontstash.h
```diff
@@ -1069,12 +1076,131 @@ static void fons__blur(FONScontext* stash, unsigned char* dst, int w, int h, int
...
+			current=dst[t-y*b+a];
+			if(prev > current){
+				dst[t-y*b+a]=prev;
+			}
+			prev=current;
+		}
+	}
+}
+
+// Gray level morphological dilation approximated by convoling with a max stencil along
```
Iniquitatis, line 1184:
```suggestion
// Gray level morphological dilation approximated by convolving with a max stencil along
```

src/fontstash.h
```diff
@@ -1088,7 +1214,9 @@ static FONSglyph* fons__getGlyph(FONScontext* stash, FONSfont* font, unsigned in
 
 	if (isize < 2) return NULL;
 	if (iblur > 20) iblur = 20;
-	pad = iblur+2;
+	if (idilate > 20) idilate = 20;
+	const int antiAliasBonus = 2;
+	pad= antiAliasBonus + iblur + idilate;
```
Iniquitatis, line 1219:
```suggestion
	pad = antiAliasBonus + iblur + idilate;
```

fwsGonzo:
Fantastic work! Thanks for doing this.

PR #641, opened 2022-07-30
Line Styles (dashed, dotted, and glow)

rgb2hsv:
This PR adds different line styles by generating uv textures along stroke paths and then applying a shading style. Extra vertices are inserted so that the UV distortion occurs only at the joint between line segments. There should be no distortion along straight paths, which is preferred for a rectangular selection tool.  
dump


rgb2hsv:
Added fixes for gles and addressed bug where line UVs get reverse when winding order is enforced.

src/nanovg.c
```diff
@@ -1398,7 +1407,7 @@ static void nvg__flattenPaths(NVGcontext* ctx)
 		}
 
 		// Enforce winding.
-		if (path->count > 2) {
+		if (state->lineStyle > 1 && path->count > 2) {
```
rgb2hsv, line 1410: Avoid flipping polygon in line style mode because this will flip UVs and cause dashes to jump around.

src/nanovg.c
```diff
@@ -1398,7 +1407,7 @@ static void nvg__flattenPaths(NVGcontext* ctx)
 		}
 
 		// Enforce winding.
-		if (path->count > 2) {
+		if (state->lineStyle == NVG_LINE_SOLID && path->count > 2) {
```
rgb2hsv, line 1410: For dotted and dashed lines we don't want to reverse the polygon because that will cause the glyphs to jump around as the curve changes shape.

PR #646, opened 2022-08-14
Fix text jitter for animated (moving) text

rgb2hsv:
Font stash returns glyph positions in integer pixels. When rendering text at floating point positions,  glyph positions get rounded to the nearest pixel. This is particularly noticeable with animated moving text, giving the appearance of jitter relative to other animated graphics. To circumvent this problem, this PR renders text at the origin (0,0) and then translates it afterwards to the proper floating point pixel location. This is possible because text rendering is shift invariant. In addition to updating the draw text call, all the methods for getting text bounds and positions have been updated as well.

dump

### DIFF
--- a/src/nanovg.c
+++ b/src/nanovg.c
@@ -132,6 +132,7 @@ struct NVGcontext {
 	int fillTriCount;
 	int strokeTriCount;
 	int textTriCount;
+	int textTextureDirty;
 };
 
 static float nvg__sqrtf(float a) { return sqrtf(a); }
@@ -163,6 +164,7 @@ static float nvg__normalize(float *x, float* y)
 	return d;
 }
 
+static void nvg__flushTextTexture(NVGcontext* ctx);
 
 static void nvg__deletePathCache(NVGpathCache* c)
 {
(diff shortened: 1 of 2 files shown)


src/nanovg.c
```diff
@@ -2487,7 +2493,7 @@ float nvgText(NVGcontext* ctx, float x, float y, const char* string, const char*
 	verts = nvg__allocTempVerts(ctx, cverts);
 	if (verts == NULL) return x;
 
-	fonsTextIterInit(ctx->fs, &iter, x*scale, y*scale, string, end, FONS_GLYPH_BITMAP_REQUIRED);
```
rgb2hsv, line 2490: Rendering at the origin means we don't have to care about the scale of the translation.

mulle-nat:
Sounds interesting, I will try this out sometime. Thanks for making these changes public.

Skytrias:
good change :+1:

alpheratz0:
Nice work! but sadly scaling is broken

rgb2hsv:
> Nice work! but sadly scaling is broken

Thanks for calling this out. I've added a fix for scaling and amended the test case to exercise it. Please let me know if it works for you.

alpheratz0:
> Thanks for calling this out. I've added a fix for scaling and amended the test case to exercise it. Please let me know if it works for you.

Thank you so much for the fast response and fix, now it works perfectly!!!!

PR #649, opened 2022-09-25
Allows setting the `NVG_MAX_STATES` number.

olliwang:
Currently, `NVG_MAX_STATES` is set to 32. But it’s not enough sometimes in my case. This commit allows defining the preferred `NVG_MAX_STATES` number.


memononen:
Thanks for the improvement!

olliwang:
Thanks for the fast merge :)

PR #669, opened 2024-03-12
Fix comparison in nvg__flattenPaths()

mkupchik:
A check `if (cache->npaths > 0)` on function entry does not make sense given the loop `for (j = 0; j < cache->npaths; j++)` below. If an entry check passes, then loop will never make at least one iteration. Fix initial check which appears to be an optimization for the case when there are no cached paths.


memononen:
Your fix is not correct. The check is there to see if the path is already
tessellated (can happen e.g. calling fill multiple times).

PR #677, opened 2026-02-19
Fix iPhone PNG color issues when loading images from memory in NanoVG

chentong-net:
## Background
When images are loaded via `nvgCreateImageMem`, iPhone PNG assets can show incorrect colors on Apple platforms.

## Root Cause
`nvgCreateImage` already enables STB iPhone PNG compatibility flags:
- `stbi_set_unpremultiply_on_load(1)`
- `stbi_convert_iphone_png_to_rgb(1)`

But `nvgCreateImageMem` did not, so file-based and memory-based loading paths behaved differently.

## Changes
- In `src/nanovg.c`, add the same two STB compatibility calls in `nvgCreateImageMem(...)` before `stbi_load_from_memory(...)`.


memononen:
thanks for the fix!

PR #680, opened 2026-03-28
Modernize cpp

rgb2hsv:


memononen:
I will not accept this. If you feel strongly about modern C++, I suggest to make a fork.

